PR #12, opened 2022-05-30
Integrate audit event feature for KBV CRI

techaman-m:
## Proposed changes
Added changes to integrate audit events CRI's. This works as a temporary fix that needs to be revisited later down the line. 

It may need some discussion around how we plan to use the `audience` parameter in the JWK

- [KBV-432](https://govukverify.atlassian.net/browse/KBV-432)


tal-nagra (approved):
Approving for now. We need to revisit to come up with a more robust implementation in the very near future

PR #18, opened 2022-06-16
[KBV-559] Expire Authorization Code

KaraokeStu:
### What changed

- Update handlers to return 403 access denied errors when access token, auth code, or sessions have expired

### Why did it change

In order to align with [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2) , we need to ensure when authorization codes are created, they have an expiry time set.

As per the RFC (above), this expiry time should have a maximum recommended TTL of 10 minutes.

Authorization codes must only be used once.
Once successfully changed for a bearer token, the authorization code must be expired.

### Issue tracking

- [KBV-559](https://govukverify.atlassian.net/browse/KBV-559)
- [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2)


accesstoken/src/test/java/uk/gov/di/ipv/cri/common/api/handler/AccessTokenHandlerTest.java
```diff
@@ -20,7 +20,7 @@
 import org.mockito.junit.jupiter.MockitoExtension;
 import software.amazon.awssdk.http.HttpStatusCode;
 import uk.gov.di.ipv.cri.common.library.error.ErrorResponse;
-import uk.gov.di.ipv.cri.common.library.exception.AccessTokenValidationException;
+import uk.gov.di.ipv.cri.common.library.exception.*;
```
tal-nagra, line 23: you can remove the asterisk import and replace with individual import statements

tal-nagra (changes requested):

tal-nagra (approved):

PR #38, opened 2022-09-20
Build common-lambdas with sam

idrop:
## Proposed changes

* Build common lambdas with AWS SAM cli tool instead of using gradle directly to build a zip
* Use `deploy.sh` to deploy to our dev environment
* Remove dependency on each dev maintaining a `samconfig.toml` file
* Remove unneeded dependency declarations like `soap` and `hibernate`

This has been tested with a deploy to the cri dev environment and a test of the session lambda.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+set -e
+
+RED="\033[1;31m"
+GREEN="\033[1;32m"
+NOCOLOR="\033[0m"
+
+stack_name="$1"
+audit_event_name_prefix="$2"
+cri_identifier="$3"
+
+if [ -z "$stack_name" ]
+then
+echo -e "😱 ${RED}stack name expected as first argument, e.g. ${GREEN}./deploy common-lambdas-my-name${NOCOLOR}"
+exit 1
+fi
+
+if [ -z "$audit_event_name_prefix" ]
+then
+  audit_event_name_prefix="/common-cri-parameters/AuditEventNamePrefix"
+fi
+
+if [ -z "$cri_identifier" ]
+then
+  cri_identifier="/common-cri-parameters/CriIdentifier"
+fi
+
+
+echo -e "👉 deploying common lambdas with:"
+echo -e "\tstack name: ${GREEN}$stack_name${NOCOLOR}"
+echo -e "\tAuditEventNamePrefix SSM key ${GREEN}$audit_event_name_prefix${NOCOLOR}"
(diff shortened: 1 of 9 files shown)


deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+if [ -z "$stack_name" ]
+then
+echo -e "😱 ${RED}stack name expected as first argument, e.g. ${GREEN}./deploy common-lambdas-my-name${NOCOLOR}"
+exit 1
+fi
+
+if [ -z "$secret_prefix" ]
+then
+  secret_prefix="common-lambdas"
+fi
```
idrop, line 20: ☝️ note the default secret prefix of `common-lambdas` which will be set on secrets in dev

jkunle (dismissed):

deploy.sh
tal-nagra: At present the template.yaml file does not have a SecretPrefix parameter.
Therefore, I think it can be removed.
However, new parameters are required for `AuditEventNamePrefix` & `CriIdentifier`

tal-nagra (changes requested):

tal-nagra:
I think the `gradlew` build should also be removed from: 
https://github.com/alphagov/di-ipv-cri-common-lambdas/blob/4defd0d0dd004626a989b95f55c363dba7b2ddef/.github/workflows/post-merge-matrix-deploy.yml#L63
As sam build is doing this work now

& I think we also don't need to specify the path to the template when invoking `sam package`:
https://github.com/alphagov/di-ipv-cri-common-lambdas/blob/4defd0d0dd004626a989b95f55c363dba7b2ddef/.github/workflows/post-merge-matrix-deploy.yml#L71

deploy.sh
idrop: @tal-nagra I've updated the script to remove the secretprefix and add optional AuditEventNamePrefix & CriIdentifier args

idrop:
> I think the `./gradlew build` should also be removed from:
> 
> https://github.com/alphagov/di-ipv-cri-common-lambdas/blob/4defd0d0dd004626a989b95f55c363dba7b2ddef/.github/workflows/post-merge-matrix-deploy.yml#L63
> 
> 
> As `sam build` is doing this work now
> & I think we also don't need to specify the path to the template when invoking `sam package`:
> 
> https://github.com/alphagov/di-ipv-cri-common-lambdas/blob/4defd0d0dd004626a989b95f55c363dba7b2ddef/.github/workflows/post-merge-matrix-deploy.yml#L71

I've removed these now @tal-nagra , thanks

tal-nagra (approved):
Approving based on the premise that this has been tested via a direct deployment to aws

PR #50, opened 2022-10-10
LIME-125: Added logging helper to session lambda and conditional response so we can decouple the back and frontend of passport when we switch to common lambdas

gds-jbentham:
### What changed

Added logging helper to session lambda and conditional response 

### Why did it change

So we can decouple the back and frontend of passport when we switch to common lambdas

### Issue tracking

- [LIME-125](https://govukverify.atlassian.net/browse/LIME-125)


merlinc:
Closing as the logging approach has changed since this PR was created

PR #51, opened 2022-10-11
Build common-lambdas with sam take 2

idrop:
Reverts alphagov/di-ipv-cri-common-lambdas#48


tal-nagra (approved):

session/src/main/java/uk/gov/di/ipv/cri/common/api/handler/SessionHandler.java
```diff
@@ -57,7 +59,11 @@ public SessionHandler() {
         this.eventProbe = new EventProbe();
         this.auditService =
                 new AuditService(
-                        SqsClient.builder().build(),
+                        SqsClient.builder()
```
MattBidewell, line 62: Is this meant to be here?

idrop: we added it in to address a code smell as part of this PR - good spot

PR #85, opened 2023-01-11
Access token git rearrange

merlinc:
### Issue tracking

### Other considerations


accesstoken2/src/app.ts
```diff
@@ -1,16 +1,75 @@
...
+
+            const bearerAccessTokenTTL = configService.getBearerAccessTokenTtl();
+            const accessTokenResponse = await this.accessTokenService.createBearerAccessToken(bearerAccessTokenTTL);
+            this.sessionService.createAccessTokenCode(sessionItem, accessTokenResponse);
+
+            return {
+                statusCode: 200,
+                body: JSON.stringify(accessTokenResponse),
+            };
+        } catch (err: any) {
```
MattBidewell, line 53: ideally we dont want `any` but i cant think of a elegant way of making this work

accesstoken2/src/lib/config-service.ts
```diff
@@ -0,0 +1,66 @@
...
+const SESSION_TABLE_NAME_KEY = "SessionTableName";
+const DEFAULT_AUTHORIZATION_CODE_TTL_IN_SECS = 600;
+const PARAMETER_PREFIX = process.env.AWS_STACK_NAME || "";
+
+export class ConfigService {
+    private readonly authorizationCodeTtlInMillis: number;
+    private readonly configEntries: Map<string, string>;
+
+    constructor(private ssmClient: SSMClient) {
+        const envAuthCodeTtl = parseInt(process.env.AUTHORIZATION_CODE_TTL || "", 10);
```
MattBidewell, line 12: should the optional value here definitely be an empty string?

accesstoken2/src/lib/config-service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    public async getRedirectUri(clientId: string) {
+        if (!clientId) {
+            throw new Error("Undefined clientId supplied");
+        }
+        return await this.getParameter(`clients/${clientId}/jwtAuthentication/redirectUri`);
+    }
+
+    public async getSessionTableName() {
+        return await this.getParameter(SESSION_TABLE_NAME_KEY);
```
MattBidewell, line 36: Few things here:

- this is returning the following in pseudo code
```
(Outside call)
          | Promise<getSessionTableName()>
                                              | Promise<getParameter()>
```

We don't NEED to have this function be async await. As we're going to be returning the promise created by `this.getParameter()`.

Which brings me to my second point:

- This function call isn't needed. 
We should be loading SessionTableNameKey on initiation as its something that's confirmed and not subject to change based on request.

instead, we should just access the configEntries map and return it from there. Which wont require a promise.

This code will actually be 'slow' as its a Promise and will be stuck at the end of an internal nodeJS queue to be executed. (it might not be noticeable but its still slower than it could be).

I would suggest changing this to be 

```
return this.configEntries.get(...);

File: accesstoken2/src/lib/config-service.ts [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public getAuthorizationCodeExpirationEpoch() {
+        // TODO: consider if this should be output in epoch seconds rather than milliseconds
```
MattBidewell, line 40: TODO

accesstoken2/src/lib/config-service.ts
```diff
@@ -0,0 +1,66 @@
+import { SSMClient, GetParameterCommand } from "@aws-sdk/client-ssm";
```
MattBidewell, line 1: THIS FILE SHOULD BE DELETED

accesstoken2/src/services/config-service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    constructor(private ssmClient: SSMClient) {
+        const envAuthCodeTtl = parseInt(process.env.AUTHORIZATION_CODE_TTL || "", 10);
+        this.authorizationCodeTtlInMillis =
+            (Number.isInteger(envAuthCodeTtl) ? envAuthCodeTtl : DEFAULT_AUTHORIZATION_CODE_TTL_IN_SECS) * 1000;
+        this.configEntries = new Map<string, string>();
+    }
+
+    public init(): Promise<string[]> {
+        const REDIRECT_URI = `clients/ipv-core-stub/jwtAuthentication/redirectUri`;
```
MattBidewell, line 20: We have one of the constant keys outside the init `SESSION_TABLE_NAME_KEY` and one inside. We should only use one pattern. I'd suggest setting this value up on line 2/3

accesstoken2/src/services/config-service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    public getRedirectUri(clientId: string) {
+        if (!clientId) {
+            throw new Error("Undefined clientId supplied");
+        }
+        return this.configEntries.get(`clients/${clientId}/jwtAuthentication/redirectUri`);
+    }
+
+    public async getSessionTableName() {
+        return await this.getParameter(SESSION_TABLE_NAME_KEY);
```
MattBidewell, line 34: Few things here:

this is returning the following in pseudo code
(Outside call)
          | Promise<getSessionTableName()>
                                              | Promise<getParameter()>
We don't NEED to have this function be async await. As we're going to be returning the promise created by this.getParameter().

Which brings me to my second point:

This function call isn't needed.
We should be loading SessionTableNameKey on initiation as its something that's confirmed and not subject to change based on request.
instead, we should just access the configEntries map and return it from there. Which wont require a promise.

This code will actually be 'slow' as its a Promise and will be stuck at the end of an internal nodeJS queue to be executed. (it might not be noticeable but its still slower than it could be).

I would suggest changing this to be

return this.configEntries.get(...);

accesstoken2/src/services/config-service.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public getAuthorizationCodeExpirationEpoch() {
+        // TODO: consider if this should be output in epoch seconds rather than milliseconds
```
MattBidewell, line 38: TODO

accesstoken2/src/services/config-service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    public async getJwtSigningAlgorithm(clientId: string) {
+        if (!clientId) {
+            throw new Error("Undefined clientId supplied");
+        }
+        return await this.getParameter(`clients/${clientId}/jwtAuthentication/authenticationAlg`);
+    }
+
+    public async getPublicSigningJwk(clientId: string) {
+        return await this.getParameter(`clients/${clientId}/jwtAuthentication/publicSigningJwkBase64`);
```
MattBidewell, line 86: async await isnt needed here. We can return the promise to the caller and let the caller handle the promise, it will make no difference, only make it slightly quicker.

accesstoken2/src/services/config-service.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public getBearerAccessTokenTtl(): number {
+        return Number(ACCESS_TOKEN_TTL);
```
MattBidewell, line 90: Generally not a fan of this. Could be a readonly public property on the class instead. But getter and setters makes it easier for Java folk to go to Typescript

accesstoken2/src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,55 @@
+import { importJWK, JWTPayload, jwtVerify } from "jose";
+import { ConfigService } from "./config-service";
+import { JWTVerifyOptions } from "jose/dist/types/jwt/verify";
+export class JwtVerifier {
+    constructor(private configService: ConfigService) {}
+
+    public async verify(
+        encodedJwt: any,
```
MattBidewell, line 8: Do we need ANY here, we should try not have any

accesstoken2/src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,55 @@
...
+        clientId: string,
+        mandatoryClaims: Set<string>,
+        expectedClaimValues: Map<string, string>,
+    ): Promise<JWTPayload> {
+        const signingPublicJwkBase64 = await this.configService.getPublicSigningJwk(clientId);
+        const signingPublicJwk = JSON.parse(Buffer.from(signingPublicJwkBase64, "base64").toString("utf8"));
+        const signingAlgorithm = await this.configService.getJwtSigningAlgorithm(clientId);
```
MattBidewell, line 15: Some minor performance improvements. 

we have two Promises which are using the same value to return something from the config service.

A promise is a blocking action, but in the world of NodeJS we can execute these 'simultaneously'...  (its not actually simultaneous, but it will put them on the queue one after the over so in theory it should be faster).

To do so we do this.

```
const signingPublicJwkBase64Promise = this.configService.getPublicSigningJwk(clientId);
const signingAlgorithmPromise = this.configService.getJwtSigningAlgorithm(clientId)
const [signingPublicJwkBase64, signingAlgorithm] = Promise.all([signingPublicJwkBase64Promise, signingAlgorithmPromise]);
```

accesstoken2/src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,55 @@
...
+        const publicKey = await importJWK(signingPublicJwk, signingPublicJwk.alg);
+
+        const jwtVerifyOptions = this.createJwtVerifyOptions(signingAlgorithm, expectedClaimValues);
+        const { payload } = await jwtVerify(encodedJwt, publicKey, jwtVerifyOptions);
+
+        if (mandatoryClaims?.size) {
```
MattBidewell, line 21: What happens if mandatoryClaims size = 0?

accesstoken2/src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,55 @@
...
+            mandatoryClaims.forEach((mandatoryClaim) => {
+                if (!payload[mandatoryClaim]) {
+                    throw new Error(`Claims-set missing mandatory claim: ${mandatoryClaim}`);
```
MattBidewell, line 24: should this be an Error or a custom Error

accesstoken2/src/services/session-service.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    public async createAccessTokenCode(sessionItem: SessionItem, accessToken: BearerAccessToken) {
+        const tableName = await this.configService.getSessionTableName();
+        //     // Expire the authorization code immediately, as it can only be used once
+        //     sessionItem.authorizationCode = '';
+        //     // Set the access token
+        //   sessionItem.accessToken = accessToken.access_token;
+        //      // Set the access token expiry
+        //   sessionItem.accessTokenExpiryDate = this.configService.getBearerAccessTokenExpirationEpoch();
+        console.log(`sessionItem ${sessionItem}, accessToken ${JSON.stringify(accessToken)}`);
```
MattBidewell, line 71: Remove console.logs

tidy up these comments are they needed?

accesstoken2/src/types/errors.ts
```diff
@@ -0,0 +1,55 @@
...
+    constructor(m: string) {
+        super(m);
+        this.statusCode = 400;
+    }
+}
+
+export class ServerError extends BaseError {
+    constructor(m: string) {
+        super("Server error");
+        this.statusCode = 500;
```
MattBidewell, line 37: we need the codes for these other errors

accesstoken2/src/lib/config-service.ts
```diff
@@ -0,0 +1,66 @@
...
+        const defaultKeys = [SESSION_TABLE_NAME_KEY];
+        const promises = [];
+        for (const key of defaultKeys) {
+            promises.push(this.getParameter(key));
+        }
+
+        return Promise.all(promises);
+    }
+
+    public async getRedirectUri(clientId: string) {
```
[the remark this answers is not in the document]
techaman-m: this is `java`-ish code

accesstoken2/src/app.ts
jkunle: Yeah, doesn't appear there is a solution for this `any` from yesterday I think any seems to be the best fit

jkunle (approved):

MattBidewell (approved):
🤞🏻

PR #87, opened 2023-01-18
Refactor accesstokenvalidator

jkunle:
## Proposed changes

Moved conditionals from main handler into Accesstokenvalidator.

### What changed

Introduce type for token request

### Why did it change

Refactored
### Issue tracking

### Other considerations

### DIFF
--- a/accesstoken2/src/app.ts
+++ b/accesstoken2/src/app.ts
@@ -9,7 +9,6 @@ import { ConfigService } from "./services/config-service";
 import { AccessTokenRequestValidator } from "./services/token-request-validator";
 import { AccessTokenService } from "./services/access-token-service";
 import { JwtVerifier } from "./services/jwt-verifier";
-import { InvalidAccessTokenError, InvalidRequestError } from "./types/errors";
 
 const logger = new Logger();
 const metrics = new Metrics();
@@ -29,40 +28,19 @@ export class AccessTokenLambda implements LambdaInterface {
     public async handler(event: APIGatewayProxyEvent, context: any): Promise<APIGatewayProxyResult> {
         try {
             await initPromise;
-
-            // validate the incoming payload
-            const requestPayload = event.body;
-            if (!requestPayload) {
-                throw new InvalidRequestError("Invalid request: missing body");
-            }
-
-            this.accessTokenValidator.validatePayload(requestPayload);
-
-            const searchParams = new URLSearchParams(requestPayload);
-            const authCode = searchParams.get("code") as string;
-            const sessionItem = await this.sessionService.getSessionByAuthorizationCode(authCode);
-
+            
+            const requestPayload = this.accessTokenValidator.validatePayload(event.body);
+            const sessionItem = await this.sessionService.getSessionByAuthorizationCode(requestPayload.code);
             logger.appendKeys({ govuk_signin_journey_id: sessionItem.clientSessionId });
 
-            this.accessTokenValidator.validateTokenRequestToRecord(
-                authCode,
-                sessionItem
-            );
+            this.accessTokenValidator.validateTokenRequestToRecord(requestPayload.code, sessionItem)
 
             const expectedAudience = await configService.getJwtAudience(sessionItem.clientId);
-            if (!expectedAudience) {
-                throw new InvalidRequestError("audience is missing");
-            }
-
-            const client_assertion = searchParams.get("client_assertion") as string
-            const jwtPayload = await this.accessTokenValidator.verifyJwtSignature(
-                Buffer.from(client_assertion, "utf-8"),
+            await this.accessTokenValidator.verifyJwtSignature(
+                Buffer.from(requestPayload.client_assertion, "utf-8"),
                 sessionItem.clientId,
                 expectedAudience
             );
-            if (!jwtPayload.jti) {
-                throw new InvalidRequestError("jti is missing");
-            }
 
             const bearerAccessTokenTTL = configService.getBearerAccessTokenTtl();
             const accessTokenResponse = await this.accessTokenService.createBearerAccessToken(bearerAccessTokenTTL);
@@ -75,14 +53,13 @@ export class AccessTokenLambda implements LambdaInterface {
         } catch (err: any) { //Todo dont want any
             logger.error({
                 statusCode: err.statusCode ?? 500,
-                message: err.message,
+                message: err?.message,
                 err: err
             });
-
             return {
                 statusCode: err.statusCode ?? 500,
                 body: JSON.stringify({
-                    message: err.statusCode >= 500 ? "Server error" : err.message,
+                    message: err?.statusCode >= 500 ? "Server Error": err.message,
                     code: err.code || null,
                     errorSummary: err.getErrorSummary(),
                 }),
--- a/accesstoken2/src/models/request_payload.ts
+++ b/accesstoken2/src/models/request_payload.ts
@@ -1,9 +1,0 @@
-export class RequestPayload {
-    constructor(
-        public grantType: string,
-        public code: string,
-        public redirectUri: string,
-        public clientAssertionType: string,
-        public clientAssertion: string
-    ) { }
-}
--- a/accesstoken2/src/models/request_payload.ts
+++ b/accesstoken2/src/models/request_payload.ts
@@ -1,0 +1,9 @@
+export class RequestPayload {
+    constructor(
+        public grantType: string,
+        public code: string,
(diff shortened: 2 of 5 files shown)


accesstoken2/src/app.ts
```diff
@@ -29,40 +29,22 @@ export class AccessTokenLambda implements LambdaInterface {
     public async handler(event: APIGatewayProxyEvent, context: any): Promise<APIGatewayProxyResult> {
         try {
             await initPromise;
-
+            const request = event.body as string
```
MattBidewell, line 32: is the event body a string?

accesstoken2/src/app.ts
```diff
@@ -72,17 +54,28 @@ export class AccessTokenLambda implements LambdaInterface {
                 statusCode: 200,
                 body: JSON.stringify(accessTokenResponse),
             };
-        } catch (err: any) { //Todo dont want any
+        } catch (err: unknown) { //Todo dont want any
```
MattBidewell, line 57: I think we're adding complexity with this catch statement, I think all we can simplify this

accesstoken2/src/app.ts
```diff
@@ -72,17 +54,28 @@ export class AccessTokenLambda implements LambdaInterface {
...
+        function logError(err: CustomErrors) {
             logger.error({
                 statusCode: err.statusCode ?? 500,
-                message: err.message,
+                message: err?.message,
                 err: err
             });
-
+        }
+        function getError(err: CustomErrors): APIGatewayProxyResult | PromiseLike<APIGatewayProxyResult> {
```
MattBidewell, line 74: we should avoid functions that have two return types that could be a promise sometimes, and sometimes are not a promise

accesstoken2/src/models/request_payload.ts
```diff
@@ -0,0 +1,9 @@
+export class RequestPayload {
```
MattBidewell, line 1: The Java in you is here 👀 

This should likely be in the `types` directory and shouldn't be a class. It should be a Type or Interface instead. I think it should be a type.

```
export type RequestPayload {
  grantType: string;
  code: string:
....
}

File: accesstoken2/src/services/token-request-validator.ts [hunk 5]
```diff
@@ -2,60 +2,55 @@ import { ConfigService } from "./config-service";
...
-        sessionItem: SessionItem,
-    ): void {
+        return new RequestPayload(grant_type, code, redirectUri, client_assertion_type, client_assertion);
+    };
 
-        if (authCode !== sessionItem.authorizationCode) {
-            throw new InvalidAccessTokenError();
-        }
+    public validateTokenRequestToRecord(authCode: string, sessionItem: SessionItem) {
+        if (!sessionItem) return new Error("Invalid sessionItem")
```
MattBidewell, line 41: Lets not throw new Error, lets throw a `InvalidPayload` error or whatever it is

accesstoken2/src/services/token-request-validator.ts
```diff
@@ -2,60 +2,55 @@ import { ConfigService } from "./config-service";
...
 
         const configRedirectUri = this.configService.getRedirectUri(sessionItem.clientId);
         if (configRedirectUri !== sessionItem.redirectUri) {
             throw new InvalidRequestError(`Invalid request: redirect uri ${sessionItem.redirectUri} does not match configuration uri ${configRedirectUri}`)
         }
     }
 
-    public async verifyJwtSignature(jwt: Buffer, clientId: string, audience: string): Promise<JWTPayload> {
-        return await this.jwtVerifier.verify(
+    public async verifyJwtSignature(jwt: Buffer, clientId: string, audience: string) {
```
MattBidewell, line 50: Lets add return type of `Promise<void>`

accesstoken2/src/types/errors.ts
```diff
@@ -1,5 +1,5 @@
 // Implementation of ErrorResponse.java in di-ipv-cri-lib
-class BaseError extends Error {
+export class BaseError extends Error {
```
MattBidewell, line 2: Ideally the BaseError should never be exported (in my opinion)
Instead, its a base class that's used by the other classes to create more specific errors relating to our system. I created the baseError so that we could add the relevant types and functions without code duplication.

accesstoken2/src/app.ts
```diff
@@ -72,17 +54,28 @@ export class AccessTokenLambda implements LambdaInterface {
                 statusCode: 200,
                 body: JSON.stringify(accessTokenResponse),
             };
-        } catch (err: any) { //Todo dont want any
+        } catch (err: unknown) { //Todo dont want any
+            for (const appError of ApplicationErrors) {
```
MattBidewell, line 58: I suspect we can use some Typescript magic to make this for of loop redundant

accesstoken2/src/app.ts
jkunle: I don't know but I have taken this onboard

```
 } catch (err: unknown) { //Todo dont want any
            logError(err as CustomErrors);
            return getError(err as CustomErrors);
        }
        function logError(err: CustomErrors) {
            logger.error({
                statusCode: err.statusCode ?? 500,
                message: err?.message,
                err: err
            });
        }
        function getError(err: CustomErrors): APIGatewayProxyResult {
            if (err == undefined || err.statusCode == undefined) return { statusCode: 500, body: JSON.stringify("Unknown Error!")}

            return {
                statusCode: err.statusCode ?? 500,
                body: JSON.stringify({
                    message: err?.statusCode >= 500 ? "Server Error": err.message,
                    code: err.code || null,
                    errorSummary: err.getErrorSummary(),
                }),
            };
        }
```
This is close to your original and seems to work

accesstoken2/src/app.ts
jkunle: I opted to

```
            const request = event.body
            // validate the incoming payload
            const requestPayload = this.accessTokenValidator.validatePayload(request);
```

```
 validatePayload(tokenRequestBody: string | null)
```

jkunle: Yeah it is a string or null

accesstoken2/src/app.ts
```diff
@@ -72,17 +54,28 @@ export class AccessTokenLambda implements LambdaInterface {
...
+                if (err instanceof appError) {
+                    logError(err);
+                    return getError(err);
+                }
+            }
+            logError(err as unknown as BaseError);
+            return Promise.reject({statusCode: 500, message: "Unknown Error Occurred"});
```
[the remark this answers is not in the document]
jkunle: I don't know what I was thinking off, I think it was the vscode refactoring when I extracted that method it added the `PromiseLike` stuff

I have removed this

accesstoken2/src/app.ts
jkunle: Yeah, seems to be fine with `APIGatewayProxyResult`

accesstoken2/src/models/request_payload.ts
jkunle: I wanted to return a new `RequestPayload` object

```
 return new RequestPayload(grant_type, code, redirectUri, client_assertion_type, client_assertion);
```

So I can keep the `URLSearchParams` out of the handler and work with it like so
```
requestPayload.code
requestPayload.clientAssertion
```

accesstoken2/src/types/errors.ts
jkunle: I have made the abstract class

accesstoken2/src/app.ts
jkunle: Yeah actually I didn't need I had two ideas in my head and implemented both

accesstoken2/src/models/request_payload.ts
MattBidewell: You can do that by doing the following with a type - 

```
const myReturnedObject: RequestPayload = {
    grantType: grant_type,
    code,
    redirectUri,
    clientAssertionType: client_assertion_type,
    clientAssertion: client_assertion
    
};
return myReturnedObject
```

accesstoken2/src/app.ts
MattBidewell: This is better but again I dont think we need a specific function to force the logger to log the error, we're adding no additional functionality by creating new functions to logError and getError

jkunle: I removed the methods, I thought it helped with the err declaration but `any` and `unknown` are pretty much in the same level

MattBidewell (approved):

PR #88, opened 2023-01-18
BAU: Remove public Api key

jkunle:
Api key is only used for non dev enviroments and not required for dev as a result it is also not required during pre-merge tes

## Proposed changes

Remove the specification of public api key as an environment variable

### What changed

Corresponding environment variable removed  

### Why did it change

Discovered during QA exploration of how to run the common lambda api from the commandline

Removed `PUBLIC_API_KEY` environment variable

### Other considerations

The github secret value associated with this value should also be removed


gokhakan (approved):

techaman-m (approved):
LGTM

PR #102, opened 2023-01-31
BAU - refactor Typescript lamdbda functions implementation

tal-nagra:
### What changed
- Removed duplicated/redundant code
- General tidy-up

### Why did it change
- The code was in an intermediate state with previous parallel work-streams duplicating classes/modules etc


jkunle (approved):

merlinc (approved):
Approved, its non-production at the moment, so let's fix forward.

PR #103, opened 2023-02-01
BAU: Added test for request utils

jkunle:
## Proposed changes

Added test for request-utils

### DIFF
--- a/lambdas/tests/unit/common/utils/request-utils.test.ts
+++ b/lambdas/tests/unit/common/utils/request-utils.test.ts
@@ -1,0 +1,80 @@
+import { jest } from '@jest/globals';
+import { APIGatewayProxyEvent } from "aws-lambda";
+import { getClientIpAddress, getSessionId } from "../../../../src/common/utils/request-utils";
+describe('request-utils', () => {
+    describe("getClientIpAddress", () => {
+        test("returns the value of x-forwarded-for header", () => {
+            const result = getClientIpAddress({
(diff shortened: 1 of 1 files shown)


lambdas/tests/unit/common/utils/request-utils.test.ts
```diff
@@ -0,0 +1,45 @@
+import { getClientIpAddress, getSessionId } from "../../../../src/common/utils/request-utils";
+
+describe("getClientIpAddress", () => {
```
merlinc, line 3: Could we wrap this in a `request-utils` block, similar to [session-service-test](https://github.com/alphagov/di-ipv-cri-common-lambdas/blob/main/lambdas/tests/unit/services/session-service.test.ts#L22)

lambdas/tests/unit/common/utils/request-utils.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    test("returns the value of x-forwarded-for header", () => {
+        const result = getClientIpAddress({
+            headers: {
+                "x-forwarded-for": "192.168.1.1"
```
merlinc, line 7: Could we get a test for different casing of the header? e.g. `X-FORWARDED-FOR` - the spec says they should be case insensitive, and we're using a `lowercase` in the code, so it'd be good to test that option too.

scronk1 (approved):

PR #104, opened 2023-02-02
BAU: Added test for jwtverifier

jkunle:
## Proposed changes

Added test coverage for jwt-verifier

### DIFF
--- a/lambdas/src/common/security/jwt-verifier.ts
+++ b/lambdas/src/common/security/jwt-verifier.ts
@@ -20,13 +20,14 @@ export class JwtVerifier {
             const jwtVerifyOptions = this.createJwtVerifyOptions(signingAlgorithm, expectedClaimValues);
             const { payload } = await jwtVerify(encodedJwt, publicKey, jwtVerifyOptions);
 
-            if (mandatoryClaims && mandatoryClaims.size) {
-                mandatoryClaims.forEach((mandatoryClaim) => {
-                    if (!payload[mandatoryClaim]) {
-                        throw new Error(`Claims-set missing mandatory claim: ${mandatoryClaim}`);
-                    }
-                });
-            }
+            if (!mandatoryClaims || mandatoryClaims?.size === 0) throw new Error("No mandatory claims provided");
+
+            mandatoryClaims?.forEach((mandatoryClaim) => {
+                if (!payload[mandatoryClaim]) {
+                    throw new Error(`Claims-set missing mandatory claim: ${mandatoryClaim}`);
+                }
+            });
+            
             return payload;
         } catch (error) {
             this.logger.error("JWT verification failed", error as Error);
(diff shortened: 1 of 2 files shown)


lambdas/tests/unit/services/jwt-verifier.test.ts
```diff
@@ -0,0 +1,211 @@
...
+import * as jose from 'jose'
+import { importJWK, JWTHeaderParameters, jwtVerify } from "jose";
+import { JwkKeyExportOptions } from "crypto";
+jest.mock("jose", () => ({
+    importJWK: jest.fn(),
+    jwtVerify: jest.fn(),
+}));
+
+describe("JwtVerifier", () => {
+    const logger = {
```
merlinc, line 13: You don't want to assign variables outside of a `before` / `beforeEach` block as you will end up with weird errors due to how the tests are instantiated

Define the variables using a `let` in the block so they can be reused, and then assign them in a `beforeEach` - that way the value is guaranteed to remain consistent between tests.

Basically, the testing framework runs all the code as it's assembling the tests. If it's normal code its run instantly, but if its Jest functions (e.g `describe` / ` beforeEach`) it'll still run instantly, but the Jest function will create a hook using the internal function to be run later

lambdas/src/common/security/jwt-verifier.ts
```diff
@@ -20,13 +20,14 @@ export class JwtVerifier {
...
             const { payload } = await jwtVerify(encodedJwt, publicKey, jwtVerifyOptions);
 
-            if (mandatoryClaims && mandatoryClaims.size) {
-                mandatoryClaims.forEach((mandatoryClaim) => {
-                    if (!payload[mandatoryClaim]) {
-                        throw new Error(`Claims-set missing mandatory claim: ${mandatoryClaim}`);
-                    }
-                });
-            }
+            if (mandatoryClaims?.size == 0) throw new Error("No mandatory claims provided");
```
merlinc, line 23: This comparison is going to potentially be `undefined == 0` - which doesn't quite do what we expect:

```
undefined == 0
false
undefined === 0
false
```
In JavaScript it's usually better to test for [strict equality ===](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality) rather than [equality ==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality), as the latter does implicit type casting.

The previous code was checking if the value was passed through as undefined, or if it had a size property (assuming its a method), which is a very JavaScript way of defensively coding.

This is Typescript, so the property has to be a Set, which means it has a `size` method, even if that's 0. I'm not 100% sure what the behaviour is when its `undefined` - maybe try writing a test to expose that and see what happens?

merlinc (changes requested):

merlinc (approved):
I'd like to restructure these tests with layered `describe` blocks - but I don't think that's a reason to hold this PR up

PR #128, opened 2023-02-15
OG-1226: token-request-validator unit test

surajkumar:
### What changed
Created unit tests for the token-request-validator. 

### Why did it change

Unit test coverage

### Issue tracking

- [OJ-1226](https://govukverify.atlassian.net/browse/OJ-1226)

### DIFF
--- a/lambdas/tests/unit/services/token-request-validator.test.ts
+++ b/lambdas/tests/unit/services/token-request-validator.test.ts
@@ -1,0 +1,168 @@
+import { AccessTokenRequestValidator } from "../../../src/services/token-request-validator";
+import { JwtVerifierFactory } from "../../../src/common/security/jwt-verifier";
+import { SessionItem } from "../../../src/types/session-item";
+describe("token-request-validator.ts", () => {
+    let accessTokenRequestValidator: AccessTokenRequestValidator;
+    const mockJwtVerifierFactory = jest.mocked(JwtVerifierFactory);
+    const code = "test";
+    const redirect_uri = "http://abc123.com";
+    const client_assertion = "test";
+    const client_assertion_type = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
+    const grant_type = "authorization_code";
+
+    beforeEach(() => {
+        jest.resetAllMocks();
+        accessTokenRequestValidator = new AccessTokenRequestValidator(mockJwtVerifierFactory.prototype);
+    });
+
+    describe("validatePayload", () => {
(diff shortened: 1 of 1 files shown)


lambdas/tests/unit/services/token-request-validator.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    beforeEach(() => {
+        jest.resetAllMocks();
+        accessTokenRequestValidator = new AccessTokenRequestValidator(mockJwtVerifierFactory.prototype);
+    });
+
+    describe("validatePayload", () => {
+        it("should throw when the client_assertion_type is not valid", function () {
+            const tokenRequestBody = `code=${code}&redirect_uri=${redirect_uri}&client_assertion=${client_assertion}&client_assertion_type=test&grant_type=${grant_type}`;
+            let requestPayload;
+            try {
```
scronk1, line 22: I generally prefer using `expect(() => accessTokenRequestValidator.validatePayload(tokenRequestBody)).toThrow("Invalid grant_type parameter")` because it avoids relying on the catch statement for your expectation to be run - if it doesn't error then the statement is missed completely

scronk1: The problem is also avoided with those expect assertions(2) statement because then the test knows it needs to hit both

lambdas/tests/unit/services/token-request-validator.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    const client_assertion_type = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
+    const grant_type = "authorization_code";
+
+    beforeEach(() => {
+        jest.resetAllMocks();
+        accessTokenRequestValidator = new AccessTokenRequestValidator(mockJwtVerifierFactory.prototype);
+    });
+
+    describe("validatePayload", () => {
+        it("should throw when the client_assertion_type is not valid", function () {
```
scronk1, line 19: Did you check that the grant_type and client_assertion_type was a mismatch in the java code too like the guys in slack recommended?

surajkumar: Was planning on leaving this as is for now since if it is a bug, it will need to be fixed in another ticket

lambdas/tests/unit/services/token-request-validator.test.ts
surajkumar: Good spot, will update

scronk1 (approved):

PR #133, opened 2023-02-17
TS lambda test coverage

rahulGDS:
### What changed

Acceptance Test coverage for TS lambda

### Why did it change

To increase the coverage for both JAVA and TS lamda

### Issue tracking

### Other considerations

### DIFF
--- a/integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
+++ b/integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
@@ -21,8 +21,10 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 public class APISteps {
 
     private static final String ENVIRONMENT = "/dev"; // dev, build, staging, integration
-    private static final String DEV_SESSION_URI = ENVIRONMENT + "/session";
-    private static final String DEV_AUTHORIZATION_URI = ENVIRONMENT + "/authorization";
+    private static String DEV_SESSION_URI = ENVIRONMENT + "/session";
+
+    private static String DEV_AUTHORIZATION_URI = ENVIRONMENT + "/authorization";
+    public static String DEV_AcessToken_URI = ENVIRONMENT + "/token";
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private static final String DEFAULT_REDIRECT_URI =
             "https://di-ipv-core-stub.london.cloudapps.digital/callback";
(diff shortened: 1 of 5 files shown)


integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
```diff
@@ -21,8 +21,10 @@
 public class APISteps {
 
     private static final String ENVIRONMENT = "/dev"; // dev, build, staging, integration
-    private static final String DEV_SESSION_URI = ENVIRONMENT + "/session";
-    private static final String DEV_AUTHORIZATION_URI = ENVIRONMENT + "/authorization";
+    private static String DEV_SESSION_URI = ENVIRONMENT + "/session";
+
+    private static String DEV_AUTHORIZATION_URI = ENVIRONMENT + "/authorization";
+    public static String DEV_AcessToken_URI = ENVIRONMENT + "/token";
```
merlinc, line 27:
```suggestion
    public static String DEV_ACCESS_TOKEN_URI = ENVIRONMENT + "/token";
    
```

merlinc: This uses the same naming standard as the existing variables

integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
```diff
@@ -40,6 +42,21 @@ public void setAuthorizationJARForTestUser(int rowNumber)
         sessionRequestBody = IpvCoreStubUtil.sendCreateSessionRequest(userIdentityJson);
     }
 
+    @Given("Session lambda implementation is in {string}")
+    public void setSessionEndpoint(String endPoint) {
+        if (endPoint.equals("TS")) DEV_SESSION_URI = ENVIRONMENT + "/session-two";
```
merlinc, line 47: Could we add some logging to the other methods, so we can be sure that we're hitting the expected endpoints?

If this code doesn't work for some reason, it will fall back to the existing Java endpoints.

integration-tests/src/test/resources/features/AccessTokenAPI.feature
```diff
@@ -1,19 +1,41 @@
 Feature: Access Token API
 
-  Background: a session id is returned
+  Scenario Outline: access token is returned
     Given authorization JAR for test user 681
+    And AccessToken lambda implementation is in '<SessionLambdaImplementation>'
```
merlinc, line 5: `Given authorization JAR for test user 681`
```suggestion
    And Session lambda implementation is in '<SessionLambdaImplementation>'
```

All these lines that are paired with `authorization JAR` should be referring to the Session lambda

merlinc (changes requested):
Can we get some logging of the URLs that are being used, so we can be sure that we're hitting the correct endpoints? The current implementation does a fallback to the Java code, so if there's bugs in our step definitions, we wont notice that we're not hitting the TypeScript code

merlinc (dismissed):
LGTM so far - but it's failing on one of the acceptance tests, so can't be merged yet

merlinc (approved):

PR #134, opened 2023-02-17
OJ-1231: use specific SessionValidationError

jkunle:
## Proposed changes

Instead of using generic Error this PR ensures the specification of a defined error type
the mirrors the java one see https://github.com/alphagov/di-ipv-cri-lib/blob/806c02d8beda586f301030b85b3a0d070a64bfd0/src/main/java/uk/gov/di/ipv/cri/common/library/error/ErrorResponse.java#L22
Instead of returning a ValidationResult and then inspecting it for which error to handle
throw the expected error and allow this to be handled by the catch block in the handler.
This aligns with pattern used in the access-token-handler

### What changed

Stopped using ValidationResult, throw specific error

### Why did it change

To ensure we handle errors, in a similar approach.

### Issue tracking

- [OJ-1231](https://govukverify.atlassian.net/browse/OJ-1231))

[OJ-1231]: https://govukverify.atlassian.net/browse/OJ-1231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/services/auth-request-validator.ts
+++ b/lambdas/src/services/auth-request-validator.ts
@@ -1,32 +1,30 @@
 import { APIGatewayProxyEventQueryStringParameters } from "aws-lambda/trigger/api-gateway-proxy";
-import { ValidationResult } from "../types/validation-result";
+import { SessionValidationError } from "../types/errors";
 
 export class AuthorizationRequestValidator {
     validate(
         queryStringParams: APIGatewayProxyEventQueryStringParameters | null,
         sessionClientId: string,
         expectedRedirectUri: string,
-    ): ValidationResult {
+    ) {
         if (!queryStringParams) {
-            return { isValid: false, errorMsg: "Missing querystring parameters" };
+            throw new SessionValidationError("Session Validation Exception", "Missing querystring parameters");
         }
 
         const clientId = queryStringParams["client_id"];
         const redirectUri = queryStringParams["redirect_uri"];
         const responseType = queryStringParams["response_type"];
-        let errorMsg = null;
         if (!clientId) {
-            errorMsg = "Missing client_id parameter";
+            throw new SessionValidationError("Session Validation Exception", "Missing client_id parameter");
         } else if (!redirectUri) {
-            errorMsg = "Missing redirect_uri parameter";
+            throw new SessionValidationError("Session Validation Exception", "Missing redirect_uri parameter");
         } else if (!responseType) {
-            errorMsg = "Missing response_type parameter";
+            throw new SessionValidationError("Session Validation Exception", "Missing response_type parameter");
         } else if (clientId !== sessionClientId) {
-            errorMsg = "Invalid client_id parameter";
+            throw new SessionValidationError("Session Validation Exception", "Invalid client_id parameter");
         } else if (redirectUri !== expectedRedirectUri) {
-            errorMsg = "Invalid redirect_uri parameter";
+            throw new SessionValidationError("Session Validation Exception", "Invalid redirect_uri parameter");
         }
 
-        return { isValid: !errorMsg, errorMsg: errorMsg };
     }
 }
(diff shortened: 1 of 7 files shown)


lambdas/src/services/auth-request-validator.ts
```diff
@@ -1,32 +1,30 @@
...
+            throw new SessionValidationError("Session Validation Exception", "Missing querystring parameters");
         }
 
         const clientId = queryStringParams["client_id"];
         const redirectUri = queryStringParams["redirect_uri"];
         const responseType = queryStringParams["response_type"];
-        let errorMsg = null;
         if (!clientId) {
-            errorMsg = "Missing client_id parameter";
+            throw new SessionValidationError("Session Validation Exception", "Missing client_id parameter");
```
surajkumar, line 18: Just a question: For the end user experience, it might be better to throw a complex error containing all the validation failures rather than them having to call the api multiple times? Less requests this way and not limiting the error to first reported error?

lambdas/src/types/errors.ts
```diff
@@ -47,9 +46,22 @@ export class JwtSignatureValidationError extends BaseError {
 }
 
 export class SessionNotFoundError extends BaseError {
-    constructor(id: string) {
+    constructor(public readonly id: string) {
         super(`Could not find session item with id: ${id}`);
         this.statusCode = 400; // check
```
surajkumar, line 51: Understand this isn't part of your change but should we do something about this "check" comment?

lambdas/tests/unit/common/utils/request-utils.test.ts
```diff
@@ -56,23 +56,22 @@ describe("request-utils", () => {
...
         });
     });
     describe("matching headers", () => {
         test("throws if there are multiple session-id headers", () => {
             jest.spyOn(global.Object, "keys").mockReturnValueOnce(["session-id", "session-id"]);
             const event = {
-                headers: {} as any,
+                headers: {} as unknown,
             } as unknown as APIGatewayProxyEvent;
             expect(() => getSessionId(event)).toThrow;
```
surajkumar, line 76: Also not specific to your changes but we should add the error message here too since it provides on.

lambdas/src/services/auth-request-validator.ts
jkunle: For now the spirit is a like for like interpretation of the way it is done in Java land and not to really to extend it in anyway for now, your suggestion is valid could be for future consideration.

surajkumar (commented):
See comments

lambdas/src/types/errors.ts
jkunle: I suppose Matt, was trying to say check that this is actually equivalent in Java land, I guess he was unsure about something, I think there is a ticket to check things typescript land are equivalent in java land so this seems to fall into that space

lambdas/tests/unit/common/utils/request-utils.test.ts
```diff
@@ -60,20 +61,24 @@ describe("request-utils", () => {
...
-            } as unknown as APIGatewayProxyEvent;
-            expect(() => getSessionId(event)).toThrow;
+        test("getSessionId throws an error if there are multiple session-id headers", () => {
+            const event: Partial<APIGatewayProxyEvent> = {
+                multiValueHeaders: {
+                    "session-id": ["123", "456"],
+                },
+            } as unknown as APIGatewayProxyEventHeaders;
+
+            expect(() => getSessionId(event as APIGatewayProxyEvent)).toThrow(new InvalidRequestError("Unexpected quantity of session-id headers encountered: 2"));
```
surajkumar, line 81: Prettier?

surajkumar (changes requested):

surajkumar (approved):
Looking good

PR #135, opened 2023-02-17
Chore/unit test coverage and file rename

scronk1:
### What changed

unit test coverage for auth handler
rename unit test file to name original ts file

### Other considerations


surajkumar (approved):
OK

PR #143, opened 2023-02-22
Oj 1257 access token handler coverage new branch on main

surajkumar:
### What changed

session-service.ts
errors.ts
access-token-handler.test.ts

### Why did it change

session-service does not check if access or auth codes have expired like in the Java versions. This PR adds that and unit tests. 

### Issue tracking

- [OJ-1257](https://govukverify.atlassian.net/browse/OJ-1257)

[OJ-1257]: https://govukverify.atlassian.net/browse/OJ-1257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/services/session-service.ts
+++ b/lambdas/src/services/session-service.ts
@@ -56,11 +56,15 @@ export class SessionService {
             throw new InvalidAccessTokenError();
         }
 
-        if (sessionItem.Items[0].authorizationCodeExpiryDate < new Date().getTime()) {
+        if (this.hasAuthCodeExpired(sessionItem.Items[0].authorizationCodeExpiryDate)) {
             throw new AuthorizationCodeExpiredError();
         }
 
         return sessionItem.Items[0] as SessionItem;
+    }
+
+    private hasAuthCodeExpired(authorizationCodeExpiryDate: number): boolean {
+        return authorizationCodeExpiryDate < Math.floor(Date.now() / 1000);
     }
 
     public async createAccessTokenCode(sessionItem: SessionItem, accessToken: BearerAccessToken) {
--- a/lambdas/src/services/session-service.ts
+++ b/lambdas/src/services/session-service.ts
@@ -3,12 +3,7 @@ import { SessionItem } from "../types/session-item";
 import { BearerAccessToken } from "../types/bearer-access-token";
 import { ConfigService } from "../common/config/config-service";
 import { randomUUID } from "crypto";
-import {
-    AccessCodeExpiredError,
-    AuthorizationCodeExpiredError,
-    InvalidAccessTokenError,
-    SessionNotFoundError,
-} from "../types/errors";
+import { AuthorizationCodeExpiredError, InvalidAccessTokenError, SessionNotFoundError } from "../types/errors";
 import { SessionRequestSummary } from "../types/session-request-summary";
 import { CommonConfigKey } from "../types/config-keys";
 
@@ -61,12 +56,8 @@ export class SessionService {
             throw new InvalidAccessTokenError();
         }
 
-        if (sessionItem.Items[0].authorizationCodeExpiryDate < new Date().getDate()) {
+        if (sessionItem.Items[0].authorizationCodeExpiryDate < new Date().getTime()) {
             throw new AuthorizationCodeExpiredError();
-        }
-
-        if (sessionItem.Items[0].accessTokenExpiryDate < new Date().getDate()) {
-            throw new AccessCodeExpiredError();
         }
 
         return sessionItem.Items[0] as SessionItem;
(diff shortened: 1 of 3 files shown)


lambdas/src/types/errors.ts
```diff
@@ -69,3 +69,19 @@ export class SessionValidationError extends BaseError {
         return `${this.code}: ${this?.details}`;
     }
 }
+
+export class AuthorizationCodeExpiredError extends BaseError {
+    constructor() {
+        super("Authorization code expired");
+        this.statusCode = 403;
+        this.code = 1026;
```
jkunle, line 77: 1027

surajkumar: Updated

lambdas/src/services/session-service.ts
```diff
@@ -56,6 +61,14 @@ export class SessionService {
             throw new InvalidAccessTokenError();
         }
 
+        if (sessionItem.Items[0].authorizationCodeExpiryDate < new Date().getDate()) {
+            throw new AuthorizationCodeExpiredError();
+        }
+
+        if (sessionItem.Items[0].accessTokenExpiryDate < new Date().getDate()) {
```
jkunle, line 68: I think this is out of scope of this ticket, it is the Credential issue lambda that needs to check whether the accesstoken has expired which is further upstream (we haven't converted issuercredential lambda that why the method appeared to be missing, it is not i.e it is the java sessionService). Whereas the Accesstoken lambda needs to know if the authorizationCode created downstream i.e created when /Authorize endpoint was hit) is still valid

lambdas/src/types/errors.ts
```diff
@@ -69,3 +69,19 @@ export class SessionValidationError extends BaseError {
...
+
+export class AuthorizationCodeExpiredError extends BaseError {
+    constructor() {
+        super("Authorization code expired");
+        this.statusCode = 403;
+        this.code = 1027;
+    }
+}
+
+export class AccessCodeExpiredError extends BaseError {
```
jkunle, line 81: Also out of scope due to the above explanation

lambdas/tests/unit/handlers/access-token-handler.test.ts
```diff
@@ -131,6 +139,109 @@ describe("access-token-handler.ts", () => {
                 expect(output.statusCode).toBe(200);
                 expect(metricsSpy).toHaveBeenCalledWith("accesstoken", MetricUnits.Count, 1);
             });
+
+            it("should return http 200 if the authorizationCodeExpiryDate is within date", async () => {
```
jkunle, line 143: This also out of scope

lambdas/tests/unit/handlers/access-token-handler.test.ts
```diff
@@ -131,6 +139,109 @@ describe("access-token-handler.ts", () => {
...
+                        },
+                    } as unknown as APIGatewayProxyEvent,
+                    null,
+                );
+
+                expect(output.statusCode).toBe(200);
+                expect(metricsSpy).toHaveBeenCalledWith("accesstoken", MetricUnits.Count, 1);
+            });
+
+            it("should return http 200 if the accessTokenExpiryDate has not expired", async () => {
```
jkunle, line 195: This is also out of scope

lambdas/tests/unit/handlers/access-token-handler.test.ts
jkunle, line 143: We already have a 200 success test, I would rather see a 403 error test instead the whole point of this ticket was to try and increase test coverage for access-token-handler. Also I think you should mock or spy of the date, rather than making a call to getDate, the test could start feel under some circumstances

surajkumar: I added a 403 test under the "fail paths" alongside a 200 test in the success paths.

jkunle:
I think the focus should more of test coverage increase here, we do have ticket to explore typescript vs java coverage we should avoid this being about that. I details the error by test for in the ticket and they of the class 403 error which there isn't coverage for

lambdas/tests/unit/handlers/access-token-handler.test.ts
surajkumar: Removed

lambdas/src/types/errors.ts
surajkumar: Removed

lambdas/src/services/session-service.ts
surajkumar: OK - Removed it

lambdas/tests/unit/handlers/access-token-handler.test.ts
jkunle: Please check the ticket details to see what I mean, we can have quick chat if it makes things clearer

surajkumar: Updated `session-service#getSessionByAuthorizationCode` to check against current epoch time.
As discussed the 403 unit test is fine but I have also added a test to check the 403 when we have multiple session items.

jkunle (approved):

PR #145, opened 2023-02-22
OJ-1256: throw specific sessionValidationerr

jkunle:
## Proposed changes

Throw specific errors instead of returning a generic error result object, 
now error handling is done in manner consistent with the other handlers

### What changed

Error handling in SessionHandler, aligned with the other handler

### Why did it change

To allow a consist way of handling errors

- [OJ-1256](https://govukverify.atlassian.net/browse/OJ-1256)

[OJ-1256]: https://govukverify.atlassian.net/browse/OJ-1256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/tests/unit/services/security/jwe-decrypter.test.ts
+++ b/lambdas/tests/unit/services/security/jwe-decrypter.test.ts
@@ -45,15 +45,11 @@ describe("JweDecrypter", () => {
         kmsClient = new KMSClient({});
         jweDecrypter = new JweDecrypter(kmsClient, getEncryptionKeyId.mockReturnValueOnce("test-key-id"));
         jest.spyOn(JSON, "parse").mockReturnValueOnce(jweProtectedHeader);
+        jest.clearAllMocks();
     });
 
     it("decrypts JWE", async () => {
         const decoded_iv = new Uint8Array([1, 2, 3, 4]);
-        const response = {
-            Plaintext: "decryptedContentEncKey",
-        };
-        const kmsClientMock = <jest.Mock>kmsClient.send;
-        kmsClientMock.mockResolvedValue(response);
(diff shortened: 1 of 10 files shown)


lambdas/tests/unit/services/security/jwe-decrypter.test.ts
```diff
@@ -63,16 +65,26 @@ describe("JweDecrypter", () => {
         expect(result).toEqual(Buffer.from("decrypted content"));
     });
 
+    it("throws error when decrypts JWE fails", async () => {
+        try {
+            await jweDecrypter.decryptJwe(compactJwe);
+        } catch (error) {
```
scronk1, line 71: Typically with these I prefer to do something like
`expect(() => jweDecrypter.decryptJwe(compactJwe)).toThrow(JweDecrypterError("..."))`
because if the error never throws the test will just end and not throw a failure

jkunle: Yeah, I used that in one of the test, but I find they are sometime limited. In this case I wanted to do a `contains` part of the error message. Generally, I don't like the try catch approach but I opted for them if it is easier or makes the test a little clearier on the asserts

scronk1: That's fair - it is really clear. The expects do stack so you can do things like `.toThrow(expect.objectContaining({...}))` if you just want to test part of it? Alternately could you just add an `expect.assertions(2)` at the end of each test just to check nothing is getting missed off

scronk1 (approved):

PR #150, opened 2023-02-24
OJ-1256: replace try catch style assertions in auth-request-validator test

jkunle:
## Proposed changes

Change assertion style

### What changed

Change assertion style from try-catch to expect-throw

### Why did it change

Recommended improvement


merlinc (approved):
Nice!

PR #152, opened 2023-02-24
OJ-1220 update and increase logging

scronk1:
### What changed

Added more descriptive logging

### Why did it change

Improve debugging

### Issue tracking

- [OJ-1220](https://govukverify.atlassian.net/browse/OJ-1220)

**Example log streams** 
Session - https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdi-ipv-cri-common-kbv-sarah-SessionFunctionTwo/log-events/2023$252F02$252F28$252F$255B$2524LATEST$255Dea5432cfc3e642a39ff34cec630107e5

Authorization - https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdi-ipv-cri-common-kbv-sarah-AuthorizationFunctionTwo/log-events/2023$252F02$252F28$252F$255B$2524LATEST$255D1ba16481f40e4dc1b5fc99e05e08676d

Access token - https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdi-ipv-cri-common-kbv-sarah-AccessTokenFunctionTwo/log-events/2023$252F02$252F28$252F$255B$2524LATEST$255Dc18cc78205c8459ea68c92e42346a8fe

### Other considerations

[OJ-1220]: https://govukverify.atlassian.net/browse/OJ-1220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/handlers/access-token-handler.ts
+++ b/lambdas/src/handlers/access-token-handler.ts
@@ -64,7 +64,7 @@ export class AccessTokenLambda implements LambdaInterface {
             const accessTokenResponse = await this.bearerAccessTokenFactory.create();
             await this.sessionService.createAccessTokenCode(sessionItem, accessTokenResponse);
 
-            logger.info("Access token created")
+            logger.info("Access token created");
 
             metrics.addMetric(ACCESS_TOKEN, MetricUnits.Count, 1);
 
--- a/lambdas/src/handlers/access-token-handler.ts
+++ b/lambdas/src/handlers/access-token-handler.ts
@@ -29,7 +29,7 @@ export class AccessTokenLambda implements LambdaInterface {
 
     @logger.injectLambdaContext({ clearState: true })
     @metrics.logMetrics({ throwOnEmptyMetrics: false, captureColdStartMetric: true })
-    public async handler(event: APIGatewayProxyEvent, context: any): Promise<APIGatewayProxyResult> {
+    public async handler(event: APIGatewayProxyEvent, _context: any): Promise<APIGatewayProxyResult> {
         try {
             await initPromise;
 
--- a/lambdas/src/handlers/access-token-handler.ts
+++ b/lambdas/src/handlers/access-token-handler.ts
@@ -75,7 +75,7 @@ export class AccessTokenLambda implements LambdaInterface {
         } catch (err: any) {
             metrics.addMetric(ACCESS_TOKEN, MetricUnits.Count, 0);
             //Todo dont want any
-            logger.error(`Access Token Lambda error occurred: ${err.getErrorSummary()}`, err as Error);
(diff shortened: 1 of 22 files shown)


jkunle (dismissed):

lambdas/src/handlers/session-handler.ts
```diff
@@ -54,6 +54,8 @@ export class SessionLambda implements LambdaInterface {
     public async handler(event: APIGatewayProxyEvent, context: any): Promise<APIGatewayProxyResult> {
         try {
             const deserialisedRequestBody = JSON.parse(event.body as string);
+            logger.info("Session lambda triggered with event", deserialisedRequestBody);
```
merlinc, line 57: What does the structure of `deserialisedRequestBody` look like, and is there any PII information in it?

lambdas/src/handlers/access-token-handler.ts
```diff
@@ -33,10 +33,12 @@ export class AccessTokenLambda implements LambdaInterface {
         try {
             await initPromise;
 
+            logger.info("Access Token Lambda triggered with event", JSON.stringify(event));
```
merlinc, line 36: What is the full set of properties in the event, do we need them all?

Also, I can't remember - does the `AWS logger` accept an object for the 2nd property, or does it need to be `Stringified`? It'd be more performant if it could remain as JSON.

lambdas/src/handlers/session-handler.ts
jkunle: Oh that's a really good point

jkunle: I don't think so, but this really points to the reason why we only log specific things, like I forgot completely about PII if we start to do this kind of logging one day something we don't want would make it through

lambdas/src/handlers/access-token-handler.ts
scronk1: It only accepts strings unfortunately, I'll strip out the bits of the event it actually uses and log those instead of the whole object

lambdas/src/handlers/session-handler.ts
scronk1: If you think there's a risk then I can strip out all the event printing and just stick to the status / progress logging

merlinc (approved):
The object logging has been removed until we find a good solution for redaction :+1: 

LGTM

PR #155, opened 2023-02-24
OJ-1270: Fix eslint warning forbidden non null assertion warnings

jkunle:
## Proposed changes

Code improvement to make ESlint happier

### Issue tracking

- [OJ-1270](https://govukverify.atlassian.net/browse/OJ-1270)

[OJ-1270]: https://govukverify.atlassian.net/browse/OJ-1270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/common/config/config-service.ts
+++ b/lambdas/src/common/config/config-service.ts
@@ -46,8 +46,11 @@ export class ConfigService {
         return this.clientConfigurations.has(clientId);
     }
 
-    public getClientConfig(clientId: string): Map<string, string> | undefined {
-        return this.clientConfigurations.get(clientId);
+    public getClientConfig(clientId: string): Map<string, string>{
+        if (!this.clientConfigurations.get(clientId)){
+            throw new Error(`no configuration for client id ${clientId}`);
+        }
+        return this.clientConfigurations.get(clientId) as Map<string, string>;
     }
 
     public getConfigEntry(key: CommonConfigKey) {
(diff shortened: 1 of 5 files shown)


lambdas/src/common/config/config-service.ts
```diff
@@ -35,10 +35,10 @@ export class ConfigService {
         if (ssmParameters.length === 0) {
             throw new Error(`No client config found. Invalid client id encountered: ${clientId}`);
         }
-        const clientConfigEntries: Map<string, string> = new Map<string, string>();
-        ssmParameters.forEach((p) => {
-            clientConfigEntries.set(p.Name!.substring(p.Name!.lastIndexOf("/") + 1), p.Value!);
-        });
+
+        const clientConfigEntries = new Map(
+            ssmParameters.map((p) => [p?.Name?.substring(p?.Name?.lastIndexOf("/") + 1), p?.Value]),
```
merlinc, line 40: The previous code had `p.Name!` which was requiring that `p.Name` not be undefined. It's also in an impure function way adding data into a variable in an enclosed scope (so in theory, could handle missing values);

The new code run this as a map, so there's a locked down 1:1 relationship on the object structure, and makes the properties optional.

I think this means we could have a mapped value of `[undefined: undefined]`. I'm not 100% sure - there's a lot of `?` in there, and it's Friday afternoon!

If we're going to make a chain of optional properties, could we get a few more tests to more clearly declare what we're expecting to happen?

merlinc (approved):
LGTM

PR #157, opened 2023-02-28
feat: non blocking switchable matrix

mrdoodles:
## Proposed changes

This is an initial implementation of the non blocking matrix which implements fail-fast: false but does not prevent the corresponding build version of a failed deployment from attempting to run which will be added in another PR. This change also introduces switching of deployments on and off via setting repository variables, the default not to run if the corresponding variable is missing or set to false.

### What changed

* The matrixes have had the fail-fast strategy set to false 
* An ENABLED matrix variable has been added for each job run which is set via repository variables

### Why did it change

These changes have been made to:
* Prevent one CRI failure blocking the deployment for others
* Enable turning off of deployments without changing the workflow so that breaking changes can be controlled  

### Issue tracking

- [OJ-1234](https://govukverify.atlassian.net/browse/OJ-1234)

- [X] Documented in the [README](./blob/main/README.md)

### Other considerations

Repository variables will need to be added to enable the deployments
 
[OJ-1234]: https://govukverify.atlassian.net/browse/OJ-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/post-merge-matrix-deploy.yml
+++ b/.github/workflows/post-merge-matrix-deploy.yml
@@ -8,29 +8,40 @@ on:
 jobs:
   publish_common_cri_to_matrix_dev:
     strategy:
+      fail-fast: false
       matrix:
-        target: [ ADDRESS_DEV, FRAUD_DEV, KBV_DEV, KBV_POC, CIC_DEV ]
+        target: [ ADDRESS_DEV, FRAUD_DEV, KBV_DEV, KBV_POC, CIC_DEV, DL_DEV ]
         include:
           - target: ADDRESS_DEV
+            ENABLED: "${{ vars.ADDRESS_DEV_ENABLED }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/post-merge-matrix-deploy.yml
```diff
@@ -41,52 +52,64 @@ jobs:
       id-token: write
       contents: read
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout code
+        if: matrix.ENABLED == 'true'
```
merlinc, line 56: Are you able to have this conditional one level above, or does that skip the entire matrix?

mrdoodles: No can't have it above as it is not defined at that point, I did try a few variations on it but got invalid workflow syntax, GitHub is a bit lacking in this area

merlinc (approved):
Annoying that we have to remember `if: matrix.ENABLED == 'true'` - but that's GHA for you

PR #159, opened 2023-03-01
Oj 1624 epoch changes in config service

surajkumar:
### What changed

config-service.ts
config-service.test.ts

### Why did it change

The current implementation uses milliseconds when we need epoch seconds to match Java land. 
config-service.test.test changes include changing the expected ms result to seconds as that's what the dynamo db table expects. 

### Issue tracking

- [OJ-1624](https://govukverify.atlassian.net/browse/OJ-1624)

### DIFF
--- a/lambdas/src/common/config/config-service.ts
+++ b/lambdas/src/common/config/config-service.ts
@@ -81,7 +81,7 @@ export class ConfigService {
 
     public getSessionExpirationEpoch() {
         const sessionTtl = parseInt(this.getConfigEntry(CommonConfigKey.SESSION_TTL), 10);
-        return Math.floor(((Date.now() + sessionTtl) * 1000) / 1000);
+        return Math.floor((Date.now() + sessionTtl * 1000) / 1000);
     }
 
     public getBearerAccessTokenTtl(): number {
--- a/lambdas/src/common/config/config-service.ts
+++ b/lambdas/src/common/config/config-service.ts
@@ -76,7 +76,7 @@ export class ConfigService {
     }
 
     public getAuthorizationCodeExpirationEpoch() {
-        return Math.floor(Date.now() + this.authorizationCodeTtlInMillis * 1000) / 1000;
+        return Math.floor(Date.now() + this.authorizationCodeTtlInMillis) / 1000;
     }
 
     public getSessionExpirationEpoch() {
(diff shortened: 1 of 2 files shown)


lambdas/src/common/config/config-service.ts
```diff
@@ -76,7 +76,7 @@ export class ConfigService {
     }
 
     public getAuthorizationCodeExpirationEpoch() {
-        return Math.floor(Date.now() + this.authorizationCodeTtlInMillis * 1000) / 1000;
+        return Math.floor(Date.now() + this.authorizationCodeTtlInMillis) / 1000;
```
surajkumar, line 79: This does not have * 1000 as that's done in the constructor already

lambdas/src/common/config/config-service.ts
```diff
@@ -76,22 +76,20 @@ export class ConfigService {
     }
 
     public getAuthorizationCodeExpirationEpoch() {
-        // TODO: consider if this should be output in epoch seconds rather than milliseconds
-        // so that it is consistent with the java implementation
-        return Date.now() + this.authorizationCodeTtlInMillis;
+        return Math.floor(Date.now() + this.authorizationCodeTtlInMillis) / 1000;
```
scronk1, line 79: I think the `Math.floor()` should apply to the whole phrase - if we're doing the floor before the /1000 we'll still get decimals that we don't want

surajkumar: Good spot - updated

scronk1 (dismissed):

lambdas/src/common/config/config-service.ts
```diff
@@ -76,22 +76,20 @@ export class ConfigService {
...
-        // TODO: consider if this should be output in epoch seconds rather than milliseconds
-        // so that it is consistent with the java implementation
-        return Date.now() + this.authorizationCodeTtlInMillis;
+        return Math.floor((Date.now() + this.authorizationCodeTtlInMillis) / 1000);
     }
 
     public getSessionExpirationEpoch() {
         const sessionTtl = this.getConfigEntry(CommonConfigKey.SESSION_TTL);
-        return Date.now() + parseInt(sessionTtl, 10) * 1000;
+        return Math.floor((Date.now() + parseInt(sessionTtl, 10) * 1000) / 1000);
```
merlinc, line 84: Could we move the `parseInt` onto the previous line? Then this line would be much more similar to the other ones, only containing the ms -> s calcuations

jkunle (approved):

PR #169, opened 2023-03-07
OJ-1258: Added scope and state checks for JWT

surajkumar:
### What changed

Updated SessionRequestValidator to check if scope is openid and that a state is present within the JWT.

### Why did it change
In the Java code, the Nimbus JOSE library threw an exception when scope was missing. As this was not a custom error that we threw, it was missed when doing the Typescript conversion. This is needed functionality. 

### Issue tracking

- [OJ-1258](https://govukverify.atlassian.net/browse/OJ-1258)

### DIFF
--- a/lambdas/src/services/session-request-validator.ts
+++ b/lambdas/src/services/session-request-validator.ts
@@ -16,7 +16,12 @@ export class SessionRequestValidator {
                 "Session Validation Exception",
                 "Invalid request: JWT validation/verification failed: JWT verification failure",
             );
-        } else if (payload.client_id !== requestBodyClientId) {
+        }
+
+        const scope = payload["scope"] as string;
+        const state = payload["state"] as string;
+
+        if (payload.client_id !== requestBodyClientId) {
             throw new SessionValidationError(
                 "Session Validation Exception",
                 `Invalid request: JWT validation/verification failed: Mismatched client_id in request body (${requestBodyClientId}) & jwt (${payload.client_id})`,
@@ -31,11 +36,15 @@ export class SessionRequestValidator {
                 "Session Validation Exception",
                 `Invalid request: JWT validation/verification failed: Redirect uri ${payload.redirect_uri} does not match configuration uri ${expectedRedirectUri}`,
             );
+        } else if (!payload.scope || scope.toLowerCase().indexOf("openid") === -1) {
+            throw new SessionValidationError("Session Validation Exception", "Invalid scope parameter");
+        } else if (!state) {
+            throw new SessionValidationError("Session Validation Exception", "Invalid state parameter");
         }
 
         return payload;
     }
-    private async verifyJwtSignature(jwt: Buffer): Promise<JWTPayload | null> {
+    public async verifyJwtSignature(jwt: Buffer): Promise<JWTPayload | null> {
         const expectedIssuer = this.validationConfig.expectedJwtIssuer;
         const expectedAudience = this.validationConfig.expectedJwtAudience;
         return await this.jwtVerifier.verify(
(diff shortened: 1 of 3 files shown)


lambdas/src/services/session-request-validator.ts
```diff
@@ -31,11 +36,15 @@ export class SessionRequestValidator {
...
+        } else if (!payload.scope || scope.toLowerCase().indexOf("openid") === -1) {
+            throw new SessionValidationError("Session Validation Exception", "Invalid scope parameter");
+        } else if (!state) {
+            throw new SessionValidationError("Session Validation Exception", "Invalid state parameter");
         }
 
         return payload;
     }
-    private async verifyJwtSignature(jwt: Buffer): Promise<JWTPayload | null> {
+    public async verifyJwtSignature(jwt: Buffer): Promise<JWTPayload | null> {
```
jkunle, line 47: This method is `private` no client actually call it. Something is not quite right when you make a method that is private public only so you can test it.

lambdas/src/services/session-request-validator.ts
```diff
@@ -31,11 +36,15 @@ export class SessionRequestValidator {
                 "Session Validation Exception",
                 `Invalid request: JWT validation/verification failed: Redirect uri ${payload.redirect_uri} does not match configuration uri ${expectedRedirectUri}`,
             );
+        } else if (!payload.scope || scope.toLowerCase().indexOf("openid") === -1) {
```
jkunle, line 39: `scope.toLowerCase().includes("openid")`

lambdas/src/services/session-request-validator.ts
surajkumar: Fixed

lambdas/src/services/session-request-validator.ts
surajkumar: Updated

lambdas/tests/unit/services/session-request-validator.test.ts
```diff
@@ -122,11 +127,136 @@ describe("session-request-validator.ts", () => {
...
+            const payload = (await sessionRequestValidator.validateJwt(
+                Buffer.from("test-jwt"),
+                client_id,
+            )) as JWTPayload;
+
+            await expect(payload).toEqual(jwtPayload);
+        });
+
+        it("should fail to validate the jwt if scope is not openid", async () => {
+            const client_id = "request-client-id";
```
jkunle, line 180: I would inline some of these variables, if they are not used more than once

lambdas/tests/unit/services/session-request-validator.test.ts
```diff
@@ -122,11 +127,136 @@ describe("session-request-validator.ts", () => {
...
+            const redirect_uri = "redirect-uri";
+            const scope = "test";
+            const state = "state";
+
+            jest.spyOn(jwtVerifier.prototype, "verify").mockReturnValue(
+                new Promise<JWTPayload | null>((res) =>
```
jkunle, line 186: This all looks good and functional, could you just make a few clean up instead of new Promise<...> do a Promise.resolve

            jest.spyOn(jwtVerifier.prototype, "verify").mockReturnValueOnce(
                Promise.resolve({
                        client_id
                        redirect_uri
                        scope
                        state
                        shared_claims: personIdentity,
                } as JWTPayload),
            );

lambdas/tests/unit/services/session-request-validator.test.ts
surajkumar: Updated

lambdas/tests/unit/services/session-request-validator.test.ts
```diff
@@ -122,11 +117,125 @@ describe("session-request-validator.ts", () => {
...
+                Promise.resolve({
+                    client_id: client_id,
+                    redirect_uri: redirect_uri,
+                    scope: scope,
+                    shared_claims: personIdentity,
+                } as JWTPayload),
+            );
+
+            await expect(async () =>
+                sessionRequestValidator.validateJwt(Buffer.from("test-jwt"), client_id),
```
jkunle, line 229: This is the only variable required all the other should have been inline the same goes for the other. But this all adequate and it doesn't have to be perfect.

jkunle (approved):

PR #170, opened 2023-03-08
fix: JWT verification issue due to JWK without alg

merlinc:
### What changed

When using the TypeScript lambdas with core-stub they work fine, but coming in through the orchestrator stub there are issues.

The error message contains to
```
TypeError: \"alg\" argument is required when \"jwk.alg\" is not present
```

This suggests there is a difference in the signing key, which is overridable by config coming in via the authenticationAlg parameter

This PR defaults to the signing key alg, and falls back to the alg defined as a parameter

### Issue tracking

- [OJ-1286](https://govukverify.atlassian.net/browse/OJ-1286)

### Other considerations

[OJ-1286]: https://govukverify.atlassian.net/browse/OJ-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/common/security/jwt-verifier.ts
+++ b/lambdas/src/common/security/jwt-verifier.ts
@@ -27,7 +27,7 @@ export class JwtVerifier {
             const signingPublicJwkBase64 = this.jwtVerifierConfig.publicSigningJwk;
             const signingAlgorithm = this.jwtVerifierConfig.jwtSigningAlgorithm;
             const signingPublicJwk = JSON.parse(Buffer.from(signingPublicJwkBase64, "base64").toString("utf8"));
-            const publicKey = await importJWK(signingPublicJwk, signingPublicJwk.alg);
+            const publicKey = await importJWK(signingPublicJwk, signingPublicJwk?.alg || signingAlgorithm);
 
             const jwtVerifyOptions = this.createJwtVerifyOptions(signingAlgorithm, expectedClaimValues);
             const { payload } = await jwtVerify(encodedJwt, publicKey, jwtVerifyOptions);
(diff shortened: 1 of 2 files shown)


lambdas/src/common/security/jwt-verifier.ts
```diff
@@ -27,7 +27,7 @@ export class JwtVerifier {
             const signingPublicJwkBase64 = this.jwtVerifierConfig.publicSigningJwk;
             const signingAlgorithm = this.jwtVerifierConfig.jwtSigningAlgorithm;
             const signingPublicJwk = JSON.parse(Buffer.from(signingPublicJwkBase64, "base64").toString("utf8"));
-            const publicKey = await importJWK(signingPublicJwk, signingPublicJwk.alg);
+            const publicKey = await importJWK(signingPublicJwk, signingPublicJwk.alg || signingAlgorithm);
```
jkunle, line 30: Maybe `signingPublicJwk?.alg`

merlinc: I think it's coming through as an undefined property anyway, but I can totally defensively code :+1:

lambdas/tests/unit/common/security/jwt-verifier.test.ts
```diff
@@ -52,6 +52,49 @@ describe("jwt-verifier.ts", () => {
...
+                };
+                importJWKMock.mockResolvedValueOnce(publicKey);
+                jwtVerifyMock.mockResolvedValueOnce({
+                    payload: jwtPayload,
+                    protectedHeader: {} as JWTHeaderParameters,
+                } as unknown as Promise<jose.JWTVerifyResult & jose.ResolvedKey>);
+
+                const payload = await jwtVerifier.verify(encodedJwt, mandatoryClaims, expectedClaimValues);
+
+                expect(payload).toEqual({
```
surajkumar, line 86: Re-use the jwtPayload here

surajkumar (approved):

PR #172, opened 2023-03-09
LIME-443 Return OAuth access_denied as an expected flow when authcode…

smsgds:
### What changed

Changed the handling of sessionItem.getAuthorizationCode() not having a value, so that it is handled as an expected flow.

### Why did it change

Currently sessionItem.getAuthorizationCode() not having a value is treated as an error due to `new AuthorizationCode(sessionItem.getAuthorizationCode())` being unable to parse the authorization code and falling into the generic exception handler.

Passport and driving license CRI's will forward users to authorisation as part of the Prove another way flow, which is not an error pathway.

Instead sessionItem.getAuthorizationCode() is now checked for null/empty and will return returned  OAuth access_denied returned with HTTP 403 set.

Added the metric EVENT_NO_AUTHORIZATION_CODE = "no_authorization_code" to allow passport continue to monitor the number of users that prove another way.

Added unit test for the sessionItem.getAuthorizationCode() path.

### Issue tracking

- [LIME-443](https://govukverify.atlassian.net/browse/LIME-443)

[LIME-443]: https://govukverify.atlassian.net/browse/LIME-443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/authorization/src/main/java/uk/gov/di/ipv/cri/common/api/handler/AuthorizationHandler.java
+++ b/authorization/src/main/java/uk/gov/di/ipv/cri/common/api/handler/AuthorizationHandler.java
@@ -8,8 +8,8 @@ import com.nimbusds.oauth2.sdk.AuthorizationCode;
 import com.nimbusds.oauth2.sdk.AuthorizationSuccessResponse;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
-import org.apache.logging.log4j.Level;
 import software.amazon.awssdk.http.HttpStatusCode;
+import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.lambda.powertools.logging.CorrelationIdPathConstants;
 import software.amazon.lambda.powertools.logging.Logging;
 import software.amazon.lambda.powertools.metrics.Metrics;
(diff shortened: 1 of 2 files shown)


gds-jbentham (approved):

authorization/src/main/java/uk/gov/di/ipv/cri/common/api/handler/AuthorizationHandler.java
```diff
@@ -68,9 +70,21 @@ public APIGatewayProxyResponseEvent handleRequest(
             SessionItem sessionItem = sessionService.getSession(sessionId);
             eventProbe
                     .addJourneyIdToLoggingContext(sessionItem.getClientSessionId())
-                    .log(Level.INFO, "found session");
+                    .log(INFO, "found session");
             // validate
             authorizationValidatorService.validate(authenticationRequest, sessionItem);
+
+            // Return access denied if there is no authcode found
+            if (StringUtils.isBlank(sessionItem.getAuthorizationCode())) {
```
jkunle, line 78: Should this code not leave inside the `authorizationValidatorService` which then throws the error that can be caught and handled as an accessDenied error

smsgds: I think there is a conflict between the behaviour of all the CRI's for what a user reaching /authorize with no authcode means in the processing flow.

For FraudCRI user don't get forwarded to authorise without an auth-code having been generated (fraud returns internal server), the oauth access_denied should not break anything if one did but it would technically be an error for that CRI.

In PassportCRI as it is live (with its own version of /authorize lambda) users can reach /authorize with-out an authcode if they decide not to fill in the form and click prove-another-way which results in the OAuth access denied - it not an error pathway as such.

DrivingLicenceCRI prove-another-way has been modelled on passports way of handling prove-another-way, but is using the common-lambdas /authorize lambda, so reaching /authorize without an authcode would not be an error also.

jkunle (approved):

PR #174, opened 2023-03-09
OJ-1274: Changed esnext to "0.17.11"

surajkumar:
### What changed
template.yaml, replace esnext to 0.17.11

### Why did it change
esnext grabs the latest version available, in the past we have had problems where when the version updates for a specific tool to the latest, it breaks our workflows.

### Issue tracking

- [OJ-1274](https://govukverify.atlassian.net/browse/OJ-1274)

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -376,7 +376,7 @@ Resources:
       BuildMethod: esbuild
       BuildProperties:
         Minify: true
-        Target: "esnext"
+        Target: "0.17.11"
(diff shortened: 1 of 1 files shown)


merlinc (changes requested):
We should always be using the even numbers of a Node release, so that'd be 16 or 18.

18 is the current LTS - https://github.com/nodejs/release#release-schedule

18 is also one of the runtimes that AWS supports - https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

merlinc (approved):

PR #189, opened 2023-04-06
OJ-1413: Add middy and update packages

jkunle:
## Proposed changes

This adds middy middleware that would be used in subsequent PR's, also update other node packages

### What changed

Will be using middy to help retain `gov_uk_journey_signing_id` that gets dropped while using logger.appendKey method

### Why did it change

`gov_uk_journey_signing_id` gets dropped while using logger.appendKey method

- [OJ-1413](https://govukverify.atlassian.net/browse/OJ-1413)

[OJ-1413]: https://govukverify.atlassian.net/browse/OJ-1413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/src/common/utils/request-utils.ts
```diff
@@ -8,7 +8,7 @@ const getHeaderValue = (event: APIGatewayProxyEvent, desiredHeader: string) => {
     );
     const matchingHeadersLength =
         matchingHeaders[0] && (event?.multiValueHeaders[matchingHeaders[0]]?.length as number);
-    if (matchingHeadersLength > 1) {
+    if (+matchingHeadersLength > 1) {
```
merlinc, line 11: what does the extra `+` do?

merlinc: Ah, it's just a package update PR - I think you've got a typo here?

merlinc: Oh, it's just TypeScript being weird.

merlinc (approved):

PR #190, opened 2023-04-06
Oj-1413: use middy in access token lambda

jkunle:
## Proposed changes

This is the 1st of a few more PR'S to go that uses middy as a way to ensure that `govuk_signin_journey_id` sticks around, when using `logger.appendkeys` since middy is middleware a series of middlewares are constructed to get the persisted id

This work uncovered a few gaps as well:

- the actual value used for the id was wrong is the session lambda, value used should be `clientSessionId` which come from  jwtPayload key of the name `govuk_signin_journey_id`
- Also we do not have the value require when retrieving the sessionItem via the `getSessionByAuthCode`
 this is b'cos we a using a global index which exposes only (sessionId, redirectUri, clientId) ideally we want to expose the clientSessionId as well to get the `govuk_signin_journey_id` updating the global index is not straight forward so this should be done as separate ticket. For now a second trip is needed to retreive the full SessionItem object

### What changed

The use of several middlewares, with the aim of ensuring journey id is available it the logs

### Why did it change

see https://govukverify.atlassian.net/browse/OJ-1413

### DIFF
--- a/lambdas/src/handlers/access-token-handler.ts
+++ b/lambdas/src/handlers/access-token-handler.ts
@@ -1,28 +1,27 @@
-import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
-import { LambdaInterface } from "@aws-lambda-powertools/commons";
-import { Metrics, MetricUnits } from "@aws-lambda-powertools/metrics";
-import { Logger } from "@aws-lambda-powertools/logger";
+import middy from "@middy/core";
+import { APIGatewayProxyEvent, APIGatewayProxyResult, Context } from "aws-lambda";
 import { SessionService } from "../services/session-service";
-import { ConfigService } from "../common/config/config-service";
 import { AccessTokenRequestValidator } from "../services/token-request-validator";
 import { JwtVerifierFactory } from "../common/security/jwt-verifier";
+import { ClientConfigKey } from "../types/config-keys";
+import { BearerAccessTokenFactory } from "../services/bearer-access-token-factory";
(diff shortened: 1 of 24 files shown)


lambdas/src/handlers/access-token-handler.ts
```diff
@@ -29,30 +30,27 @@ export class AccessTokenLambda implements LambdaInterface {
...
         try {
-            await initPromise;
             logger.info("Access Token Lambda triggered");
-
-            const requestPayload = this.requestValidator.validatePayload(event.body);
-            const sessionItem = await this.sessionService.getSessionByAuthorizationCode(requestPayload.code);
-            logger.info("Session found");
+            const eventBody = event.body;
+            const sessionItem = eventBody as unknown as SessionItem;
+            const requestPayload = eventBody as unknown as RequestPayload;
```
scronk1, line 40: Is it worth doing something like `const item = eventBody as unknown as NewType` where NewType is a combo of SessionItem and RequestPayload

scronk1: I'm not 100% that it'll work but casting the event as two different types feels a bit hacky

scronk1: Like you've used below `const event_body = request.event.body as SessionItem & RequestPayload;`

jkunle: I did what you proposed initially, but decided against it. Because it's then no longer obvious which fields come from where when looking at the lambda handler code. I tested the above and it worked fine.

scronk1 (approved):

PR #211, opened 2023-05-05
BAU: Use logger instead of System.out.println and no Optional on ENV

surajkumar:
### What changed
APISteps.java

### Why did it change
* Use the Logger instead of `System.out.println` for better clarity of logs.
* Removed the use of `Optional.ofNullable` around the envs as `System.getenv()` will never be null so this call is redundant. Additionally, this removes the fallback of the PaaS URls which will eventually be decommissioned. 

### Issue tracking

- [OJ-1540](https://govukverify.atlassian.net/browse/OJ-1540)

[OJ-1540]: https://govukverify.atlassian.net/browse/OJ-1540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
+++ b/integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
@@ -32,7 +32,7 @@ public class APISteps {
             (REDIRECT_URI.toLowerCase().startsWith("http"))
                     ? REDIRECT_URI + "/callback"
                     : "https://di-ipv-core-stub.london.cloudapps.digital/callback";
-    private static final String DEFAULT_CLIENT_ID = System.getenv("DEFAULT_CLIENT_ID");
+    private static final String DEFAULT_CLIENT_ID = "null";
     private String currentAuthorizationCode;
     private String sessionRequestBody;
     private String currentSessionId;
--- a/integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
+++ b/integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
@@ -32,7 +32,7 @@ public class APISteps {
             (REDIRECT_URI.toLowerCase().startsWith("http"))
                     ? REDIRECT_URI + "/callback"
                     : "https://di-ipv-core-stub.london.cloudapps.digital/callback";
-    private static final String DEFAULT_CLIENT_ID = "null";
+    private static final String DEFAULT_CLIENT_ID = System.getenv("DEFAULT_CLIENT_ID");
     private String currentAuthorizationCode;
     private String sessionRequestBody;
     private String currentSessionId;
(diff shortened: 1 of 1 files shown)


integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
```diff
@@ -13,25 +13,23 @@
...
     private static final String ENVIRONMENT = "/dev"; // dev, build, staging, integration
     private static String devSessionUri;
     private static String devAuthorizationUri;
     public static String devAccessTokenUri;
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private static final String DEFAULT_REDIRECT_URI =
-            Optional.ofNullable(System.getenv("IPV_CORE_STUB_URL") + "/callback")
-                    .orElse("https://di-ipv-core-stub.london.cloudapps.digital/callback");
-    private static final String DEFAULT_CLIENT_ID =
-            Optional.ofNullable(System.getenv("DEFAULT_CLIENT_ID")).orElse("ipv-core-stub");
```
jkunle, line 34: We shouldn't do this yet, it save to leave this till eventual PAAS decommissioning, right now changes so far are good with KBV & ADDRESS other CRI's haven't caught up yet.

surajkumar: Actually Ola, `Optional.ofNullable(System.getenv("IPV_CORE_STUB_URL") + "/callback").orElse("https://di-ipv-core-stub.london.cloudapps.digital/callback");` will become "/callback" if `IPV_CORE_STUB_URL` env is not present. The behaviour we want of it using the default value never have worked as  `System.getenv` will never return `NULL` even if the environment variable is missing. Hence this update. 

You can test this by doing:

`String val = System.getenv("someRandomEnvThatDoesNotExit");`
You can observe that `val` in this example is not null and has the value `""`

surajkumar: What we could do instead would be something like:

`private static final String DEFAULT_REDIRECT_URI = System.getenv("IPV_CORE_STUB_URL").length() > 0 
  ? (System.getenv("IPV_CORE_STUB_URL")+ "/callback") 
  : "https://di-ipv-core-stub.london.cloudapps.digital/callback";`

integration-tests/src/test/java/gov/uk/di/ipv/cri/common/api/stepDefinitions/APISteps.java
```diff
@@ -13,25 +13,26 @@
...
+    private static final String REDIRECT_URI = System.getenv("IPV_CORE_STUB_URL");
     private static final String DEFAULT_REDIRECT_URI =
-            Optional.ofNullable(System.getenv("IPV_CORE_STUB_URL") + "/callback")
-                    .orElse("https://di-ipv-core-stub.london.cloudapps.digital/callback");
-    private static final String DEFAULT_CLIENT_ID =
-            Optional.ofNullable(System.getenv("DEFAULT_CLIENT_ID")).orElse("ipv-core-stub");
+            (REDIRECT_URI.toLowerCase().startsWith("http"))
+                    ? REDIRECT_URI + "/callback"
+                    : "https://di-ipv-core-stub.london.cloudapps.digital/callback";
+    private static final String DEFAULT_CLIENT_ID = System.getenv("DEFAULT_CLIENT_ID");
```
jkunle, line 35: Please could we put this back to `Optional.ofNullable(System.getenv("DEFAULT_CLIENT_ID")).orElse("ipv-core-stub");`

surajkumar: Done

jkunle (approved):

PR #263, opened 2024-01-11
OJ-2209: Add SonarCloud and lambda workflows

CharlesIC:
_⚠️  https://github.com/govuk-one-login/github-actions/pull/11 needs to be merged before this PR_

**OJ-2209: Add SonarCloud and lambda workflows**

- Add a workflow to run the SonarCloud and CodeQL scans on the repo and PRs
- Add the necessary config to gradle to run a SonarCloud scan

- Add a workflow to deploy to dev from a branch
  - Use a common prefix and the branch name to name the stack. 
  - Use a known set of tags so the stacks can be automatically cleaned up based on the values of those tags.

- Add a workflow to run pre-merge checks 
  The workflow deploys a preview stack and runs unit and integration tests, and other checks

- Add a workflow to delete a stack from dev 
  The workflow runs to automatically delete a stack associated with the branch when a PR is closed.

- Add a workflow to delete old preview stacks
  Filter stacks to delete based on their name and tags.

- Add a reusable integration tests workflow
  Refactor the integration tests workflow to work with the per-branch model of stack deployment.
  The new workflow uses shared GitHub Actions to reduce duplication.

**OJ-2209: Set maximum preview stack name length limit**

The common stack name is used in lambda names in the format `<common-stack-name>-<lambda-name-suffix>`. CloudFormation allows a maximum stack name length of 128 characters but the lambda service allows a maximum lambda name length of 64 characters. Therefore, the stacks we deploy must have a maximum length of `64 - <longest lambda name suffix> - 1` (1 char for the hyphen).

The longest fixed lambda name prefix we use is `AuditEventConsumerFunction` (26 chars), therefore, the stack name must not exceed 37 characters to meet this arbitrary constraint.

---

- Address linting errors

- Adjust coverage thresholds to allow jest to pass

- Run `npm install` and commit the lock file

- Delete unused `accesstoken2` project as it was causing gralde failures when running the SonarCloud scan in GHA

- Adjust the additional eslint config to only include config files in lambdas root
  It shouldn't include all the files in the lambdas directory, only the root config files.

- Update .secrets.baseline to exclude specific secrets and keywords


infrastructure/lambda/template.yaml
```diff
@@ -1354,7 +1347,6 @@ Resources:
     Type: AWS::Serverless::Function
     Condition: IsDevEnvironment
     Properties:
-      FunctionName: !Sub "${AWS::StackName}-CreateAuthCodeFunction"
       CodeUri: ../../lambdas
       Handler: create-auth-code-handler.lambdaHandler
       Runtime: nodejs18.x
```
jkunle, line 1352: Typically we should let aws create the names but there is a reason why this was the exception, all CRI's have a [private-api yaml](https://github.com/govuk-one-login/ipv-cri-kbv-api/blob/main/infrastructure/lambda/private-api.yaml#L36) referencing these names directly. Or have you figured out an alternate way this can be done?

CharlesIC: Thanks, Ola, this is really useful.

We should be using outputs from the common stack to get the names and ARNs of functions instead of fixed names because 1) it's a special knowledge you need to have in order to not break things (and I wasn't aware of this) and 2) it places an arbitrary constraint on the length of the stack name.

The correct solution is to change all CRIs to use the outputs but that's a bigger piece of work so for now I'll introduce a length limit on the stack name which has to be `64 - (<longest fixed lambda name> + 1)` - 64 is the maximum lambda name length, plus one character for the hyphen. This is really terrible but for now we have to do this.

merlinc: @CharlesIC The maximum is actually a lot less than in many cases, sometimes it's as little as 23 characters!

@mrdoodles knows more about that as he's had battles with infrastructure before

.secrets.baseline
```diff
@@ -41,7 +41,7 @@
     },
     {
       "name": "KeywordDetector",
-      "keyword_exclude": ""
+      "keyword_exclude": "secretsmanager|privateKeyFile|ARTIFACT_SOURCE_BUCKET_NAME_SECRET|GH_ACTIONS_ROLE_ARN_SECRET"
```
merlinc, line 44: I liked this approach, but I didn't see anything that allowed us to specify these values in a config file - so they're difficult to retain on the next usage of `detect-secrets scan > .secrets.baseline`

infrastructure/lambda/template.yaml
CharlesIC: I have come across this and in the past I had to set the limit to 20 something chars but in this case the limit is imposed by our own templates - setting it to 37 is enough to deploy a preview stack and run integration tests as we don't use secure pipelines for this - in cases where we do use the pipelines the limit might indeed be even less but that's a different workflow not related to this PR, thankfully.

CharlesIC: We should add comments to the template above those `FunctionName` lines to say they should not be changed or removed, since this is not obvious without the knowledge of all the CRIs.

CharlesIC: I've raised https://github.com/govuk-one-login/github-actions/pull/11 and pushed [this commit](https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/263/commits/b749db683c83d6bd6d3694bafd109e9a71de3d04) to set the lmit for integration test stacks.

.secrets.baseline
CharlesIC: The command should be run as `detect-secrets scan --baseline .secrets.baseline` - this is how pre-commit does it and it tells the `detect-secrets` utility to use the config from the baseline file - then all those keywords and other exclusions are retained.

merlinc: Ahhhh. We've been using it wrong! 

Can you update the `README.md` to mention that - it'll stop misuse in the future.

mrdoodles: Why don't we drop using the word secret, why do we need this ... what is secretsmanager ? we don't have one

lambdas/package.json
```diff
@@ -1,5 +1,16 @@
 {
     "name": "di-ipv-common-lambdas",
+    "scripts": {
```
merlinc, line 3: This has been sliced out into #265 and should be removed from this PR

.secrets.baseline
CharlesIC: Done ✅

lambdas/jest.config.ts
```diff
@@ -9,10 +9,10 @@ export default {
     setupFiles: ["<rootDir>/setEnvVars.js"],
     coverageThreshold: {
         global: {
-            statements: 95,
-            branches: 95,
+            statements: 94,
```
merlinc, line 12: This change shouldn't be needed anymore, now that you've removed the code changes.

lambdas/package.json
```diff
@@ -1,5 +1,16 @@
 {
```
merlinc, line 1: `package-lock.json` changes shouldn't be needed as you haven't installed any new packages, and `npm ci` runs correctly.

.secrets.baseline
CharlesIC: @mrdoodles these are the names of various variables/params that are in our GitHub Actions workflows and the SAM templates - they are not secret values and they're already public in this repo.

CharlesIC: I agree `ARTIFACT_SOURCE_BUCKET_NAME_SECRET ` is a bad param name but updating secret names is not in scope of this PR

lambdas/tsconfig.eslint.json
```diff
@@ -3,5 +3,5 @@
         "noEmit": true
     },
     // Additional paths to lint not included by other tsconfig.json files
```
merlinc, line 5: Comments are not valid in JSON files. If you want comments, [change the configuration file format](https://eslint.org/docs/latest/use/configure/configuration-files)

merlinc (dismissed):

lambdas/package.json
CharlesIC: The lock files needs to be commited to the repo, though. You shouldn't get a diff when running `npm install` locally - it means some dependencies are out-of-date.

lambdas/tsconfig.eslint.json
CharlesIC: This is out of scope for this PR - I changed the filter but the file already exists in this form so we can change the format (and in all the other repos) separately

lambdas/package.json
CharlesIC: This is to move the scripts section above depdendencies and make it consistent with other repos - this was not included in the smaller PR

lambdas/jest.config.ts
CharlesIC: It is still needed - I've tested just now and jest fails with the old thresholds - see [this run](https://github.com/govuk-one-login/ipv-cri-common-lambdas/actions/runs/7501550989/job/20422416611)

lambdas/tsconfig.eslint.json
merlinc: Ah, sorry - I misread this and thought you'd added that line in.

Separate PR!

jkunle (approved):

surajkumar (approved):

lambdas/jest.config.ts
merlinc: Ah damn. Ok.

PR #273, opened 2024-02-15
LIME-1019 - Updating common_lib version

ElliotMurphyGDS:
### What changed
Updating Common_lib version to be latest. This is to pull in latest release which includes updates to the audit event timestamps so that they include milliseconds.

Updated build.gradle to make nimbus versioning inline with common-lib, some fixes where required as a part of this which can be seen in the KMSRSADecrypter.java

Additional Note:
will be reviewing other releases in between 1.4.3->1.5.6 to confirm no changes will be breaking in the jump in version

### DIFF
--- a/session/src/test/java/uk/gov/di/ipv/cri/common/api/service/SignedJWTBuilder.java
+++ b/session/src/test/java/uk/gov/di/ipv/cri/common/api/service/SignedJWTBuilder.java
@@ -1,5 +1,7 @@
 package uk.gov.di.ipv.cri.common.api.service;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.JWSHeader;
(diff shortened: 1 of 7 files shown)


surajkumar (commented):
Your tests are failing because `1.5.6` hasn't been published yet. 

Could you do a test with `1.5.5` to  make sure nothing is going to break unrelated to your changes?

infrastructure/lambda/template.yaml
```diff
@@ -325,14 +325,14 @@ Mappings:
 
   IPVCoreStubAwsBuildRedirectURIMapping:
     Environment:
-      dev: "https://cri.core.build.stubs.account.gov.uk/callback"
+      dev: "http://localhost:8085/callback"
```
surajkumar: What's the reasoning for changing the URL here?

session/src/test/java/uk/gov/di/ipv/cri/common/api/service/SignedJWTBuilder.java
```diff
@@ -173,6 +178,10 @@ SignedJWT build() {
             return signedJWT;
         } catch (JOSEException e) {
             throw new IllegalStateException(e);
+        } catch (JsonMappingException e) {
```
surajkumar, line 181: Instead of 2 distinct catch blocks, you can condense this into one i.e. `} catch (JsonMappingException | JsonProcessingException e) {`

surajkumar: We should also be careful, that PII is not going to be exposed in this exception

surajkumar (changes requested):

surajkumar (approved):

gds-jbentham (approved):

gds-jbentham (approved):

merlinc (approved):

PR #295, opened 2024-05-08
LIME-1079 - Saving Nino from shared_claims to personIdentityTable for…

ElliotMurphyGDS:
### What changed
 - Storing nino value in sharedClaims sent by IPV Core
 - Mapping nino value in sharedClaims to personIdentityTable
 - Save Nino to personIdentityTable

### DIFF
--- a/lambdas/src/services/person-identity-service.ts
+++ b/lambdas/src/services/person-identity-service.ts
@@ -1,12 +1,13 @@
 import { DynamoDBDocument, PutCommand } from "@aws-sdk/lib-dynamodb";
 import { ConfigService } from "../common/config/config-service";
 import { CommonConfigKey } from "../types/config-keys";
-import { Address, BirthDate, Name, PersonIdentity } from "../types/person-identity";
+import { Address, BirthDate, Name, PersonIdentity, SocialSecurityRecord } from "../types/person-identity";
 import {
     PersonIdentityAddress,
     PersonIdentityDateOfBirth,
     PersonIdentityItem,
     PersonIdentityName,
+    PersonIdentitySocialSecurityRecord,
 } from "../types/person-identity-item";
 
 export class PersonIdentityService {
@@ -40,6 +41,7 @@ export class PersonIdentityService {
             birthDates: this.mapBirthDates(sharedClaims.birthDate),
             expiryDate: sessionExpirationEpoch,
             names: this.mapNames(sharedClaims.name),
+            socialSecurityRecord: this.mapNino(sharedClaims.socialSecurityRecord),
         };
     }
     private mapAddresses(addresses: Address[]): PersonIdentityAddress[] {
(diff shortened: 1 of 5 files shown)


lambdas/tests/unit/handlers/session-handler.test.ts
```diff
@@ -49,6 +49,11 @@ describe("SessionLambda", () => {
     let sessionRequestValidatorFactory: jest.MockedObjectDeep<typeof SessionRequestValidatorFactory>;
 
     const mockPerson: PersonIdentity = {
+        socialSecurityRecord: [
```
gds-jbentham, line 52: Do we have a test for what happens if the nino isnt in the claimset?

lambdas/tests/unit/handlers/session-handler.test.ts
```diff
@@ -49,6 +49,57 @@ describe("SessionLambda", () => {
...
+                addressLocality: "N/A",
+                postalCode: "AA1 1AA",
+                addressCountry: "UK",
+                validFrom: "2022-01",
+                validUntil: "2023-01",
+            },
+        ],
+    };
+
+    const mockPersonNoSocialSecurityRecord: PersonIdentity = {
```
gds-jbentham, line 97: This says no social record but its still present

lambdas/tests/unit/handlers/session-handler.test.ts
```diff
@@ -487,4 +538,92 @@ describe("SessionLambda", () => {
...
+                        shared_claims: mockPersonNoSocialSecurityRecord,
+                        evidence_requested: {
+                            scoringPolicy: "gpg45",
+                            strengthScore: 2,
+                        },
+                    } as JWTPayload),
+                ),
+            );
+        });
+        it("should send start audit event with context scope identity_check given a request with evidence_requested an no socialSecurityRecord", async () => {
```
gds-jbentham, line 607: We need to assert an here that nothing was saved to dynamo rather than an event

lambdas/src/services/person-identity-service.ts
```diff
@@ -40,6 +41,7 @@ export class PersonIdentityService {
             birthDates: this.mapBirthDates(sharedClaims.birthDate),
             expiryDate: sessionExpirationEpoch,
             names: this.mapNames(sharedClaims.name),
+            socialSecurityRecord: this.mapNino(sharedClaims.socialSecurityRecord),
```
Caitlin-cooling, line 44: Just thinking that this is always going to be an empty array in the NINo person details table, do you think we could optionally add it instead?

Caitlin-cooling: Oh actually think I might be getting a bit confused with the different scenarios covered in the tests so we have

1. in person-identity-service.test `should avoid formatting blank identities` `socialSecurityRecord` is being passed in as an empty array and results in an empty array
2. in session-handler.test `should save to the session with no socialSecurityRecord included` `socialSecurityRecord` is absent 

I'm guessing that if the value is empty core will not send `socialSecurityRecord` at all in the shared claims, rather than an empty array is that correct?

gds-jbentham: Yeah Core should only be sending fields required for each CRI

Caitlin-cooling (approved):
Approving from an orange point of view - the TS looks sensible and like it won't effect us 👍

PR #301, opened 2024-05-22
LIME-1055 adding mockTxmaResources value as true to deploy script

gds-adhenderson:
### What changed

Setting the value to true.

### Why did it change

So that we see the TXMA events in sqs.

### Issue tracking

- [LIME-1055](https://govukverify.atlassian.net/browse/LIME-1055)

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -31,5 +31,6 @@ sam deploy --stack-name "$stack_name" \
   cri:deployment-source=manual \
   --parameter-overrides \
   Environment=dev \
+  CreateMockTxmaResources=true \
   ${audit_event_name_prefix:+AuditEventNamePrefix=$audit_event_name_prefix} \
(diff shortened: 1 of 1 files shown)


Caitlin-cooling:
I have raised an almost identical PR here https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/299/files we chose to only add to the deploy-branch workflow because we only needed the stacks for pre-marge checks. 

Do Lime need them for post merge too?

smsgds:
> I have raised an almost identical PR here https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/299/files we chose to only add to the deploy-branch workflow because we only needed the stacks for pre-marge checks.
> 
> Do Lime need them for post merge too?

Just spotted this PR sitting here, this change to the deploy.sh would be needed for deploying dev stacks of the common-cri-api with the mocks in place. We then manually configure the template to redirect to a local stub for testing.

PR #305, opened 2024-05-31
OJ-2546: Remove mock TxMA queue resources from the SAM template

CharlesIC:
These are created by the TxMA stack in the common-infra repo instead (govuk-one-login/ipv-cri-common-infrastructure#92).

- Use the prefixed TxMA exports as the non-prefixed versions are deprecated
- Don't set default parameter values in the deploy script when those values are already the defaults in the template


kiejgordon (approved):
LGTM

jkunle (approved):

MarieseMikely (approved):

PR #309, opened 2024-06-19
OJ-2641: Support multiple scores for cri

mikebeeby:
### What changed

- Added the use of an optional CRIEvidenceProperties parameter to validate evidence requested values. 
- Removed the strengthScore parameter for all CRIs (now included in the CRIEvidenceProperties)
- Added verificationScore validation to the session-request-validator
- Modified config service parameter retrieval to only log if the CRIEvidenceProperties parameter is not found
- Added tests for the verificationScore validation. 
- Refactor some of the tests

### Why did it change

I want the system to support a Low Confidence journey for kbv by validating a verification score of 1 &  2 in the EvidenceRequest from ipv-core.

### Issue tracking

- [OJ-2641](https://govukverify.atlassian.net/browse/OJ-2641)

### Other considerations

[OJ-2641]: https://govukverify.atlassian.net/browse/OJ-2641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/common/config/config-service.ts
+++ b/lambdas/src/common/config/config-service.ts
@@ -3,6 +3,7 @@ import { Parameter } from "aws-sdk/clients/ssm";
 import { CriAuditConfig } from "../../types/cri-audit-config";
 import { ClientConfigKey, CommonConfigKey } from "../../types/config-keys";
 import { SSMProvider } from "@aws-lambda-powertools/parameters/ssm";
+import { logger } from "../utils/power-tool";
 
 const AWS_STACK_NAME_PREFIX = process.env.AWS_STACK_NAME || "";
 const AUTHORIZATION_CODE_TTL = parseNumber(process.env.AUTHORIZATION_CODE_TTL) || 600;
@@ -34,22 +35,22 @@ export class ConfigService {
         if (ssmParameters.length === 0) {
             throw new Error(`No client config found. Invalid client id encountered: ${clientId}`);
         }
-        await this.getParametersByAbsolutePath(ssmParameters, clientId);
+        await this.setParametersByAbsolutePath(ssmParameters, clientId);
     }
 
     public async initConfigWithCriIdentifierInPath(clientId: string, parameterPrefix: string, paramNameSuffix: string) {
-        const ssmParameters = await this.getParameters([
-            `/${AWS_STACK_NAME_PREFIX}/${parameterPrefix}/${paramNameSuffix}`,
-        ]);
+        const ssmParameters = await this.getCriIdentifierParameters([`/${parameterPrefix}/${paramNameSuffix}`]);
         if (ssmParameters.length === 0) {
-            throw new Error(`Invalid parameter beginning with ${parameterPrefix} encountered`);
+            logger.info(`Invalid parameter beginning with ${parameterPrefix} encountered`);
         }
-        await this.getParametersByAbsolutePath(ssmParameters, clientId);
+        await this.setParametersByAbsolutePath(ssmParameters, clientId);
     }
 
-    private async getParametersByAbsolutePath(ssmParameters: Parameter[], identifier: string) {
+    private async setParametersByAbsolutePath(ssmParameters: Parameter[], identifier: string) {
         const clientConfigEntries: Map<string, string> =
             this.clientConfigurations.get(identifier) || new Map<string, string>();
+
+        if (ssmParameters.length === 0) return;
 
         ssmParameters.forEach(({ Name, Value }) => {
             clientConfigEntries.set(...this.validateNameSuffix(Name, Value));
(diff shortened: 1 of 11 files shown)


lambdas/src/common/config/config-service.ts
```diff
@@ -34,23 +35,23 @@ export class ConfigService {
...
-        await this.getParametersByAbsolutePath(ssmParameters, clientId);
+        await this.setParametersByAbsolutePath(ssmParameters, clientId);
     }
 
-    private async getParametersByAbsolutePath(ssmParameters: Parameter[], identifier: string) {
+    private async setParametersByAbsolutePath(ssmParameters: Parameter[], identifier: string) {
         const clientConfigEntries: Map<string, string> =
             this.clientConfigurations.get(identifier) || new Map<string, string>();
 
+        if (ssmParameters.length === 0) return;
```
Caitlin-cooling, line 53: Sorry I'm not sure I'm following this so -
1. Line 35 we are checking if we have no ssm params and throwing an error if that's true
2. Line 38 we must have ssmParameters but then we are setting the parameters by absolute path which seems to actually set the clientConfigurations?
3. Line 53 we then check again to see if we have ssm params?

lambdas/src/services/session-request-validator.ts
```diff
@@ -5,16 +5,16 @@ import { ClientConfigKey, ConfigKey } from "../types/config-keys";
...
     constructor(
         private validationConfig: SessionRequestValidationConfig,
         private jwtVerifier: JwtVerifier,
-        private evidenceRequest?: EvidenceRequestConfig,
+        private criEvidenceProperties?: CRIEvidenceProperties,
     ) {}
     async validateJwt(jwt: Buffer, requestBodyClientId: string): Promise<JWTPayload> {
         const expectedRedirectUri = this.validationConfig.expectedJwtRedirectUri;
-        const configuredStrengthScore = this.evidenceRequest?.strengthScore;
+        const configuredStrengthScore = Number(this.criEvidenceProperties?.strengthScore);
```
Caitlin-cooling, line 17: Why are we doing `Number()`? The type defines `strengthScore` as an optional number, so we know that it's a number if we have it and if we don't this will produce `NaN`?

lambdas/src/services/session-request-validator.ts
```diff
@@ -27,17 +27,22 @@ export class SessionRequestValidator {
         const evidenceRequested = payload["evidence_requested"] as EvidenceRequest;
         const state = payload["state"] as string;
 
-        if (evidenceRequested && evidenceRequested.scoringPolicy !== "gpg45") {
+        if (evidenceRequested && evidenceRequested?.scoringPolicy !== "gpg45") {
```
Caitlin-cooling, line 30: [nit] why do we need the optional chain? We have already confirmed that `evidenceRequested` is truthy in the first half of this statement, no?

lambdas/src/services/session-request-validator.ts
```diff
@@ -59,6 +64,24 @@ export class SessionRequestValidator {
 
         return payload;
     }
+    private validateStrengthScore(configuredStrengthScore: number, evidenceRequested: EvidenceRequest) {
```
Caitlin-cooling, line 67: [nit] I would expect a function called `validateStrengthScore` to do some sort of validation action rather than just return true if not valid. Maybe we could rename to `isInvalid`? Or better yet, `isValid` and with `else if(!isValid())`

Caitlin-cooling: Same comment applies for `validateVerificationScore`.

lambdas/tests/unit/services/session-request-validator.test.ts
```diff
@@ -26,7 +27,7 @@ describe("session-request-validator.ts", () => {
         });
 
         it("should return an error on JWT verification failure", async () => {
-            jest.spyOn(jwtVerifier.prototype, "verify").mockReturnValue(Promise.resolve(null));
+            jest.spyOn(jwtVerifier.prototype, "verify").mockResolvedValue(null);
```
Caitlin-cooling, line 30: Lovely

lambdas/tests/unit/services/session-request-validator.test.ts
```diff
@@ -297,6 +292,167 @@ describe("session-request-validator.ts", () => {
...
+            } as SessionRequestValidationConfig;
+
+            sessionRequestValidator = new SessionRequestValidator(
+                sessionRequestValidationConfig,
+                jwtVerifier.prototype,
+                { verificationScore: [1, 2] } as CRIEvidenceProperties,
+            );
+        });
+
+        it("should pass when verificationScore is 1 and cri is di-ipv-cri-kbv-api", async () => {
```
Caitlin-cooling, line 312: Why is the CRI important here? Because its needed for the SSM param path to compare the allowed evidence requested values against the evidence requested included in the session request?

mikebeeby: Good spot, it was just from C&Ping other tests. I have removed it from these tests.

lambdas/src/common/config/config-service.ts
mikebeeby: Hmm, I have moved the return to be specifically in the initConfigWithCriIdentifierInPath function and removed the additional length check from line 53. Agreed this class is slightly confusing still though, are you happy with just this change or does it need refactoring further in this PR?

Caitlin-cooling: Thanks for removing the additional check - happy for any additional refactoring to fall outside of the scope of this PR.

What is your thinking for changing the `throw` to a `log` + `return`?  Would that mean our `initialiseClientConfigMiddleware` can still continue even if we know we're not going to be able to validate our evidence requested in the `validateJwtMiddleware`?

mikebeeby: The evidence-requested ssm param is optional, as we do not want to break other CRIs by enforcing it to be there.

This previously threw an error because we added the strengthScore ssm param to all CRIs in the CF template. Now we have removed that and used terraform to deploy evidence-requested only to the CRI's we want. 

Currently evidence-properties is the only potential parameter passed into initConfigWithCriIdentifierInPath.

mikebeeby: The alternative was to set the evidence-properties ssm param in the CF template for all CRI's and not to use terraform to deploy it. Because we have used terraform we need to handle it not being there.

Caitlin-cooling: Ah ok and the checks in `validateJwtMiddleware` now check if we have a `configuredStrengthScore` _and_ if the `evidenceRequested` matches that. Ok cool that makes sense 👍

PR #286, opened 2024-04-05
Update dependabot.yml to include gradle scanning

angold5:
Update dependabot.yml to include gradle scanning

### Issue tracking

- [OJ-XXXX](https://govukverify.atlassian.net/browse/OJ-XXX)

### Environment variables or secrets

- [ ] No environment variables or secrets were added or changed

- [ ] Documented in the [README](./blob/main/README.md)
- [ ] Added to deployment repository
- [ ] Added to local startup repository

### Other considerations

- [ ] Update [README](./blob/main/README.md) with any new instructions or tasks

PR #311, opened 2024-06-26
Adds gradle scanning to dependabot

Caitlin-cooling:
### What changed

Adds gradle scanning to dependabot as per the [recommended implementation from team manual](https://team-manual.account.gov.uk/development-standards-processes/coding-practices-and-processes/dependabot/#meeting-development-standards)

Note this is a copy of PR https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/286

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
       - dependabot
     commit-message:
       prefix: BAU
+  - package-ecosystem: "gradle" 
+    directory: "/" 
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -9,3 +9,7 @@ updates:
       - dependabot
     commit-message:
       prefix: BAU
+  - package-ecosystem: "gradle" 
+    directory: "/" 
+    schedule:
+      interval: "daily"
```
surajkumar, line 15: is there a requirement for this to be daily otherwise can we make it weekly so it's consistent with everything else please? thanks

Caitlin-cooling: Good point - done

PR #349, opened 2024-09-13
OJ-2651: pact token impl

jkunle:
## Proposed changes:

Implemented access-token-router.ts logic, drives the PACT typescript /token endpoint test
which uses the access-token-lambda factory, that employs the the `MockSSMProvider` and
`MockDynamoDBDocument` to stub out aws calls

The express application that exposes the /token endpoint needs
to be more generic as it would be used by all CRI's that consumes
the typescript token endpoint

Implement access-token-router which mocks out parameters using
`MockSSMProvider` that allows the `getParametersByName` to return values
just as the actual `SSMProvider` would but without making aws calls

`MockDynamoDBDocument` that allows mocking dynamo send and query operations
with the expected structure holding `sessionItemInDbBeforeTokenRequest`
for both outputs of send and query respectively.

The `componentId` needs to be overridden depending on the CRI, this is achieved
by using requestFilters in the pact verifier in conjunction with stateHandlers of
the verifier. The override is set as an header value that is retrieved in the incoming 
request middleware and used in access token router.

- [OJ-2651](https://govukverify.atlassian.net/browse/OJ-2651)

[OJ-2651]: https://govukverify.atlassian.net/browse/OJ-2651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/tests/unit/handlers/contract/access-token-provider.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            req.headers["component-id"] = componentId;
+            // This should not be required, but we were seeing a strange bug where pact was not generating the request with the correct content-length
+            req.headers["content-length"] = JSON.stringify(req.body).length.toString();
+            next();
+        },
+        providerVersion: "3.0.0",
+        logLevel: "info",
+    };
+
+    it("tests against potential new contracts", async () => {
```
Caitlin-cooling, line 59: What do we mean by "potential new contracts"?

lambdas/tests/unit/handlers/contract/routes/access-token-router.ts
```diff
@@ -26,23 +30,68 @@ class AccessTokenRouter {
...
-            const event = res.locals.event;
+            const body = res.locals.event.body;
+            const bodyAsJson = res.locals.json;
+            const componentId = res.locals.componentId;
 
-            const clientConfig = new Map<string, string>();
-            clientConfig.set("code", this.code);
-            clientConfig.set("redirectUri", this.redirectUri);
+            const lambda = CreateAccessTokenLambda(
+                new MockSSMProvider({
```
Caitlin-cooling, line 40: Would we expect different AccessTokenLambdas to have different ssm params? I'm interestied in why we do this here rather than in the `CreateAccessTokenLambda` function

jkunle: It was originally, in `CreateAccessTokenLambda` but the `componentId` is different depending on the CRI, so we refactor this

Caitlin-cooling: Nice like that change 👍

Caitlin-cooling (approved):
Nice work 🚀

PR #361, opened 2024-10-18
OJ-2815: Update TS session-handler to store context

mikebeeby:
### What changed

context can now be included as in the JWT for a session. This now gets context from the JWT and stores it within the session item.

### Why did it change

We want to make the corresponding changes as in the Java lambdas to store context from the JWT.

Note: check-hmrc determines the "context" of a record_check or an identity_check through the presence of the evidence_requested property. This PR allows an actual context property to be included in the JWT and stored in the session item. If this property is used in the future and wants to be included in the audit event, check-hmrc "context" logic may need refactoring. 

### Issue tracking

- [OJ-2815](https://govukverify.atlassian.net/browse/OJ-2815)

## Evidence

Tested with a check-hmrc stack through the core stub.

Screenshot 2024-10-18 at 16 40 31

[OJ-2815]: https://govukverify.atlassian.net/browse/OJ-2815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/tests/unit/services/session-service.test.ts
+++ b/lambdas/tests/unit/services/session-service.test.ts
@@ -5,6 +5,8 @@ import { InvalidAccessTokenError, SessionNotFoundError } from "../../../src/comm
 import { SessionItem } from "../../../src/types/session-item";
 import { SSMProvider } from "@aws-lambda-powertools/parameters/ssm";
 
+const UUID_REGEX = new RegExp(/^[0-9A-F]{8}-[0-9A-F]{4}-[4][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i);
+
 jest.mock("@aws-sdk/lib-dynamodb", () => {
     return {
         __esModule: true,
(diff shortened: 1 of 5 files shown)


lambdas/tests/unit/services/session-service.test.ts
```diff
@@ -218,5 +218,51 @@ describe("session-service", () => {
...
+                            redirectUri: "test-redirect-uri",
+                            state: "test-state",
+                            subject: "test-sub",
+                            context: "test-context",
+                        }),
+                    }),
+                }),
+            );
+
+            expect(output.length).toEqual(36);
```
jkunle, line 265: Why is this significant? what does 36 mean?

lambdas/tests/unit/services/session-service.test.ts
```diff
@@ -218,5 +218,51 @@ describe("session-service", () => {
 
             expect(output.length).toEqual(36);
```
jkunle, line 219: ???

lambdas/tests/unit/services/session-service.test.ts
mikebeeby: Output is the session id which is a random UUID. 36 is the length of a UUID. I guess the alternative would be to validate the UUID with regex? 

Although, yes, I did just copy this from the above test.

mikebeeby: I've pushed up a change to make it a regex check to be clearer

lambdas/tests/unit/services/session-service.test.ts
mikebeeby: Same as above

lambdas/tests/unit/services/session-service.test.ts
```diff
@@ -216,7 +218,53 @@ describe("session-service", () => {
...
+                            redirectUri: "test-redirect-uri",
+                            state: "test-state",
+                            subject: "test-sub",
+                            context: "test-context",
+                        }),
+                    }),
+                }),
+            );
+
+            expect(UUID_REGEX.test(output)).toBe(true);
```
jkunle, line 267: That's a good idea, i think `expect(output).toEqual(expect.stringMatching(UUID_REGEX));` is better

Caitlin-cooling (approved):
Nice, can we please redeploy check HMRC once this is merged to make sure that we're up to date? (let me know if you need me to flip the flags)

PR #414, opened 2025-03-24
OJ-2331: Add start endpoint to headless-core-stub

mikebeeby:
### What changed

- Added `/start` endpoint to the test-resources stack. 

### Why did it change

We want a headless, typescript core stub that has a /start endpoint that returns a JWT that can be used to start a session in any CRI frontend.

### Issue tracking

- [OJ-2331](https://govukverify.atlassian.net/browse/OJ-2331)

[OJ-2331]: https://govukverify.atlassian.net/browse/OJ-2331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test-resources/headless-core-stub/README.md
+++ b/test-resources/headless-core-stub/README.md
@@ -1,1 +1,106 @@
 # Headless Core Stub
+
+## /start
+
+This endpoint will generate an encrypted JWT that can be used to start a session in a CRI. You can pass in a JSON body to override the values of the JWT Claims Set, or if you pass in an empty JSON object it will generate with default values. 
+
+A full example of top level field overrides can be seen below. For shared_claims and evidence_requested there are more nested fields you can provide. 
+
+It is recommended to not provide overwrites for most fields. For example, time based fields - These should only be overridden if you want to test how a CRI handles expired JWTs etc.
+
+```
+{
+    "iss": "https://localhost.gov.uk",
+    "sub": "urn:fdc:gov.uk:abcdevg",
+    "aud": "https://review-a.dev.account.gov.uk",
+    "iat": 1,
+    "exp": 2,
+    "nbf": 1,
+    "response_type": "code",
+    "client_id": "ipv-core-stub-aws-headless",
+    "redirect_uri": "https://localhost.gov.uk/callback",
+    "state": "abc",
+    "govuk_signin_journey_id": "abc",
+    "shared_claims": {
+        "address": [
+            {
+                "addressLocality": "LONDON",
+                "buildingNumber": "10",
+                "postalCode": "SW1A 2AA",
+                "streetName": "DOWNING STREET",
+                "validFrom": "2020-01-01"
+            }
+        ],
+        "birthDate": [
+            {
+                "value": "1990-01-01"
+            }
+        ],
+        "name": [
+            {
+                "nameParts": [
+                    {
+                        "type": "GivenName",
+                        "value": "JOE"
+                    },
+                    {
+                        "type": "FamilyName",
+                        "value": "BLOGGS"
+                    }
+                ]
+            }
+        ]
+    },
+    "evidence_requested": {
+        "scoringPolicy": "gpg45",
+        "strengthScore": 2,
+        "verificationScore": 2
+    },
+    "context": "cri_context"
+}
+```
+
+If shared_claims is not overridden, the default will be;
+
+```
+{
+    name: [
+        {
+            nameParts: [
+                {
+                    type: "GivenName",
+                    value: "KENNETH",
+                },
+                {
+                    type: "FamilyName",
+                    value: "DECERQUEIRA",
+                },
+            ],
+        },
+    ],
+    birthDate: [
+        {
+            value: "1965-07-08",
+        },
+    ],
+    address: [
+        {
+            buildingNumber: "8",
+            streetName: "HADLEY ROAD",
+            addressLocality: "BATH",
+            postalCode: "BA2 5AA",
+            validFrom: "2021-01-01",
+        },
+    ],
+};
+```
+### Configuration
+
+This stack will need to be deployed into an account with a 'core-infrastructure' stack, as it requires the `core-infrastructure-CriDecryptionKey1Id`
+
+It requires a (test) JWK private key as stored in an SSM param at `/test-resources/ipv-core-stub-aws-headless/privateSigningKey`
+
+If you are planning to use default values for `aud`, `iss`, `redirect_uri`, all of these will need SSM parameters at;   
+`/${COMMON_LAMBDAS_STACK_NAME}/clients/ipv-core-stub-aws-headless/jwtAuthentication/audience`  
+`/${COMMON_LAMBDAS_STACK_NAME}/clients/ipv-core-stub-aws-headless/jwtAuthentication/issuer`   
+`/${COMMON_LAMBDAS_STACK_NAME}/clients/ipv-core-stub-aws-headless/jwtAuthentication/redirectUri`   
--- a/test-resources/headless-core-stub/lambdas/start/src/errors/error-response.ts
+++ b/test-resources/headless-core-stub/lambdas/start/src/errors/error-response.ts
@@ -1,0 +1,18 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { HeadlessCoreStubError } from "./headless-core-stub-error";
+
+export const handleErrorResponse = (err: unknown, logger: Logger) => {
+    err instanceof Error ? logger.error(err.message, err) : logger.error("Unknown error caught");
+
+    if (err instanceof HeadlessCoreStubError) {
+        return {
+            statusCode: err.status,
+            body: JSON.stringify({ message: err.message }),
+        };
+    }
+
+    return {
+        statusCode: 500,
+        body: JSON.stringify({ message: "Server error" }),
+    };
+};
--- a/test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
+++ b/test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
@@ -1,0 +1,110 @@
+import { IssuerAuthorizationRequestSchema } from "@govuk-one-login/data-vocab-schemas";
+import { IssuerAuthorizationRequestClass } from "@govuk-one-login/data-vocab/credentials";
+import Ajv from "ajv";
+import addFormats from "ajv-formats";
+import { v4 as uuidv4 } from "uuid";
+import { HeadlessCoreStubError } from "../errors/headless-core-stub-error";
+import { ClaimsSetOverrides } from "../types/claims-set-overrides";
+import { JWTClaimsSet } from "../types/jwt-claims-set";
+import { logger } from "../start-handler";
+
+export const parseJwtClaimsSetOverrides = (body: string | null): ClaimsSetOverrides => {
+    try {
+        const claimsSetOverrides = body ? JSON.parse(body) : {};
+        if (!claimsSetOverrides.client_id) {
+            claimsSetOverrides.client_id = "ipv-core-stub-aws-headless";
+        }
+        return claimsSetOverrides;
+    } catch (e) {
+        throw new HeadlessCoreStubError("Body is not valid JSON", 400);
+    }
+};
+
+export const generateJwtClaimsSet = async (overrides: ClaimsSetOverrides, ssmParameters: Record<string, string>) => {
+    const audience = overrides.aud || ssmParameters["audience"];
+    const issuer = overrides.iss || ssmParameters["issuer"];
+    const redirectUri = overrides.redirect_uri || ssmParameters["redirectUri"];
+
+    const now = Date.now();
+    return {
+        iss: issuer,
+        sub: overrides.sub || "urn:fdc:gov.uk:" + uuidv4(),
+        aud: audience,
+        iat: overrides.iat || msToSeconds(now),
+        exp: overrides.exp || msToSeconds(now + 5 * 60 * 1000),
+        nbf: overrides.nbf || msToSeconds(now - 1),
+        response_type: overrides.response_type || "code",
+        client_id: overrides.client_id,
+        redirect_uri: redirectUri,
+        state: overrides.state || uuidv4(),
+        govuk_signin_journey_id: overrides.govuk_signin_journey_id || uuidv4(),
+        shared_claims: overrides.shared_claims != null ? overrides.shared_claims : defaultClaims,
+        ...(overrides.evidence_requested && { evidence_requested: overrides.evidence_requested }),
+        ...(overrides.context && { context: overrides.context }),
+    } as JWTClaimsSet;
+};
+
+export const validateClaimsSet = (claimsSet: JWTClaimsSet) => {
+    //Current data-vocab schemas do not exactly match what our CRIs so we have to validate context manually and add scope and nonce
+    const claimsSetCopy = { ...claimsSet };
+    if (claimsSetCopy.context) {
+        if (typeof claimsSetCopy.context !== "string") {
+            logger.error("Invalid context field");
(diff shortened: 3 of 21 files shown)


test-resources/headless-core-stub/README.md
Caitlin-cooling, line 1: Really nice!

test-resources/headless-core-stub/README.md
```diff
@@ -1 +1,105 @@
 # Headless Core Stub
+
+## /start
+
+This endpoint will generate an encrypted JWT that can be used to start a session in a CRI.
+
+You can pass in a JSON body to override the values of the JWT Claims Set, or it will generate default values. 
+
+The aud (audience) field is required in the body. An example for Address CRI.
```
Caitlin-cooling, line 9: could we just add a bit about what this field is for?

test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
```diff
@@ -0,0 +1,106 @@
...
+const logger = new Logger();
+
+const COMMON_LAMBDAS_STACK_NAME = process.env.COMMON_LAMBDAS_STACK_NAME || "common-cri-api";
+
+const ISSUER_SSM_NAME = `/${COMMON_LAMBDAS_STACK_NAME}/clients/ipv-core-stub-aws-headless/jwtAuthentication/issuer`;
+const REDIRECT_URI_SSM_NAME = `/${COMMON_LAMBDAS_STACK_NAME}/clients/ipv-core-stub-aws-headless/jwtAuthentication/redirectUri`;
+
+export const parseJwtClaimsSetOverrides = (body: string | null): ClaimsSetOverrides => {
+    try {
+        const jwtClaimsSetOverrides = body !== null ? JSON.parse(body) : null;
```
Caitlin-cooling, line 21: [nit] could we flip this if statement so we're not having to think about double negatives?

test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
```diff
@@ -0,0 +1,106 @@
...
+        state: overrides.state || uuidv4(),
+        govuk_signin_journey_id: overrides.govuk_signin_journey_id || uuidv4(),
+        shared_claims: overrides.shared_claims != null ? overrides.shared_claims : defaultClaims,
+        ...(overrides.evidence_requested && { evidence_requested: overrides.evidence_requested }),
+        ...(overrides.context && { context: overrides.context }),
+    } as JWTClaimsSet;
+};
+
+export const validateClaimsSet = (claimsSet: JWTClaimsSet) => {
+    //Current data-vocab schemas do not exactly match what our CRIs so we have to validate context manually and add scope and nonce
```
Caitlin-cooling, line 55: Nice

test-resources/headless-core-stub/lambdas/start/src/start-handler.ts
```diff
@@ -1,11 +1,42 @@
...
+import { Logger } from "@aws-lambda-powertools/logger";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import { JWK, JWTPayload, KeyLike } from "jose";
+import { handleErrorResponse } from "./errors/error-response";
+import { generateJwtClaimsSet, parseJwtClaimsSetOverrides, validateClaimsSet } from "./services/jwt-claims-set-service";
+import { encryptSignedJwt, getPrivateSigningKey, getPublicEncryptionKey, signJwt } from "./services/signing-service";
+import { ClaimsSetOverrides } from "./types/claims-set-overrides";
+import { JWTClaimsSet } from "./types/jwt-claims-set";
+
+const logger = new Logger();
```
Caitlin-cooling, line 11: Could we add [logger context](https://docs.powertools.aws.dev/lambda/typescript/1.4.1/core/logger/#inject-lambda-context) just in case please

test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
```diff
@@ -0,0 +1,61 @@
...
+        throw new HeadlessCoreStubError("Unable to retrieve public encryption key", 500);
+    }
+
+    const base64PublicKey = Buffer.from(data.PublicKey).toString("base64");
+    const header = "-----BEGIN PUBLIC KEY-----";
+    const footer = "-----END PUBLIC KEY-----";
+    const value = base64PublicKey.match(/.{1,64}/g)?.join("\n");
+    const publicKeyPem = `${header}\n${value}\n${footer}`;
+
+    logger.error(publicKeyPem);
```
Caitlin-cooling, line 51: Probably don't need this logger?

test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    cachedPublicKey = await importSPKI(publicKeyPem, "RS256");
```
Caitlin-cooling, line 53: I'm interested in the decision to use jose and therefore need to format into pem vs using [node subtle crypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey) like the [kiwi implementation of a similar stub](https://github.com/govuk-one-login/ipv-cri-bav-api/blob/e9d98801540eeff5856d665ed3666e8a8077e860/bav-ipv-stub/src/handlers/startBavCheck.ts#L134).

I know they are getting key documents from a jwks endpoint rather than kms, but this will be our future state so wondering if this might save us some rework down the line? Appending these header and footers just feels a bit odd

test-resources/headless-core-stub/lambdas/start/src/services/ssm-service.ts
```diff
@@ -0,0 +1,31 @@
...
+        if (value) {
+            return value;
+        }
+    } catch (e) {
+        logger.debug(`Error retrieving ${name}: ` + (e as Error).stack);
+    }
+    throw new HeadlessCoreStubError(`Error retrieving ${name}`, 500);
+};
+
+export const getJsonSSMParameter = async (name: string) => {
```
Caitlin-cooling, line 21: Do we definitely need a whole separate function for this or could we have an optional `transform` parameter for the function?

test-resources/headless-core-stub/lambdas/start/src/errors/error-response.ts
```diff
@@ -0,0 +1,21 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { HeadlessCoreStubError } from "./headless-core-stub-error";
+
+export const handleErrorResponse = (err: unknown, logger: Logger) => {
+    if (err instanceof Error) {
+        logger.error(err.message, err);
```
Caitlin-cooling, line 6: Do you think there might be something helpful that we can log even if the `err` isn't an `Error`?

test-resources/headless-core-stub/lambdas/start/src/errors/error-response.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+    if (!(err instanceof HeadlessCoreStubError)) {
```
Caitlin-cooling, line 8: I worry that we would need to remember to update this if we started adding new error types. What do you think about sometihng like just 
```
const message = err.message || "Server error";
return {
    statusCode: err.status || 500,
    body: JSON.stringify({ message }),
}
```

I haven't seen this idea of using different error types to control error handling in TypeScript so much, but if you've got some docs/prior art please share!

test-resources/infrastructure/template.yaml
```diff
@@ -50,7 +54,7 @@ Globals:
   Function:
     CodeUri: ..
     Timeout: 30
-    Runtime: nodejs18.x
+    Runtime: nodejs22.x
```
Caitlin-cooling, line 57: Thanks

test-resources/infrastructure/template.yaml
```diff
@@ -137,6 +141,51 @@ Resources:
...
+            Action: 'kms:GetPublicKey'
+            Resource: !Sub 'arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/*'
+        - Statement:
+            Effect: Allow
+            Action: 'ssm:GetParameter'
+            Resource: !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/test-resources/ipv-core-stub-aws-headless/privateSigningKey"
+        - Statement:
+            Effect: Allow
+            Action: 'ssm:GetParameter'
+            Resource: !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${CommonLambdasStackName}/clients/ipv-core-stub-aws-headless/jwtAuthentication/issuer"
```
Caitlin-cooling, line 171: It's probably safe to `*` all the `ipv-core-stub-aws-headless`?

test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
mikebeeby: This logic has slightly changed with the above, should have no double negatives now

test-resources/headless-core-stub/lambdas/start/src/services/ssm-service.ts
mikebeeby: I did try this but it kind of makes the transform pointless as you still do not know the type returned. I'll have another look

test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
mikebeeby: Yeah it's not ideal but I guess it's nicer imo to use the same library for this stuff and keep it consistent with what we've used before. 
Once it's a JWK it should be simpler as we can just change to jose's `importJWK` and not have to append headers or footers.

test-resources/headless-core-stub/lambdas/start/src/types/jwt-claims-set.ts
```diff
@@ -0,0 +1,18 @@
+import { EvidenceRequestedClass, PersonExtendedMatchingClass } from "@govuk-one-login/data-vocab/credentials";
+
+export type JWTClaimsSet = {
```
surajkumar, line 3: Why do we need 2 variations? `ClaimsSetOverrides`?

test-resources/headless-core-stub/lambdas/start/tests/test-data.ts
```diff
@@ -0,0 +1,66 @@
+import { JWTPayload } from "jose";
+import { JWTClaimsSet } from "../src/types/jwt-claims-set";
+
+export class TestData {
+    static privateSigningKey = {
```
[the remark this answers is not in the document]
mikebeeby: It's from here: https://github.com/govuk-one-login/stub-oauth-client/blob/4052c908dc3a1e2b680eefa1b2bbe2d24e01593e/test/test_keys.ts#L2

test-resources/headless-core-stub/lambdas/start/src/types/jwt-claims-set.ts
mikebeeby: Ones for overriding the fields and is optional, ones the actual claims set so needs to be present

test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
```diff
@@ -0,0 +1,103 @@
...
+import { v4 as uuidv4 } from "uuid";
+import { HeadlessCoreStubError } from "../errors/headless-core-stub-error";
+import { ClaimsSetOverrides } from "../types/claims-set-overrides";
+import { getSSMParameter } from "./ssm-service";
+import { JWTClaimsSet } from "../types/jwt-claims-set";
+import { logger } from "../start-handler";
+
+const COMMON_LAMBDAS_STACK_NAME = process.env.COMMON_LAMBDAS_STACK_NAME || "common-cri-api";
+
+const AUDIENCE_SSM_NAME = `/${COMMON_LAMBDAS_STACK_NAME}/clients/ipv-core-stub-aws-headless/jwtAuthentication/audience`;
```
surajkumar, line 14: Can we pass the client_id from the incoming request and use that instead of hard coding `ipv-core-stub-aws-headless` - this would allow for scenarios such as starting from headless but finishing at core-stub or allowing for custom clients that we don't know about.

surajkumar (changes requested):

test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
mikebeeby: I think you could achieve that with just using the overrides in the existing request body?

surajkumar: Will be worth aligning with Ola, as he has this in his PR: https://github.com/govuk-one-login/ipv-cri-common-lambdas/blob/b9b82c8260735f3a00d45d325759cc40a49e25ad/test-resources/headless-core-stub/lambdas/callback/src/services/configuration-helper.ts

surajkumar: At a perf perspective, if the values are present in the body, we won't need to query SSM so keep that in mind.

test-resources/headless-core-stub/lambdas/start/src/types/jwt-claims-set.ts
surajkumar: I feel like we can just use the `ClaimsSetOverrides` in both scenarios, named as `JWTClaimsSet` - I don't understand the need for the duplication and we can express the non-optional better in the code?

surajkumar: Perhaps we can use https://zod.dev/ for this? 

```
import { z } from "zod";

const ClaimsSetOverridesSchema = z.object({
    iss: z.string({ required_error: "Issuer (iss) is required" }),
    sub: z.string({ required_error: "Subject (sub) is required" }),
    aud: z.string({ required_error: "Audience (aud) is required" }),
    iat: z.number().optional(),
    exp: z.number().optional(),
    nbf: z.number().optional(),
    response_type: z.string().optional(),
    client_id: z.string().optional(),
    redirect_uri: z.string().optional(),
    state: z.string().optional(),
    govuk_signin_journey_id: z.string().optional(),
    shared_claims: z.any().optional(), 
    evidence_requested: z.any().optional(),
    context: z.string().optional(),
});

function validateClaims(claims: ClaimsSetOverrides) {
    try {
        ClaimsSetOverridesSchema.parse(claims);
        console.log("Validation passed!");
    } catch (error) {
        console.error("Validation failed:", error);
    }
}
```

Something like this, if needed?

test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
Caitlin-cooling: Resolving as discussed

test-resources/headless-core-stub/lambdas/start/src/services/ssm-service.ts
mikebeeby: This file no longer exists

test-resources/headless-core-stub/lambdas/start/src/services/jwt-claims-set-service.ts
mikebeeby: Now uses Ola's code to get parameters

surajkumar (approved):
LGTM!

Caitlin-cooling (approved):
Think there is a small amount of tidy up to do like we have two `msToSeconds` functions now but happy for us to get this merged and come back to it later.

PR #416, opened 2025-03-24
OJ-2983: headless callback

jkunle:
## Proposed changes

Test by making a journey [this](https://govukverify.atlassian.net/wiki/spaces/OJ/pages/3717038091/Making+manual+Curl+requests+to+a+CRI) is useful reference might also be able to make a journey against check-hmrc as it has `/test/privateSigningKey`

- create a JAR [see](https://github.com/govuk-one-login/stub-oauth-client/blob/main/src/cri-crypto-utils/README.md) 
- go to the private-api /session endpoint in AWS i.e. address-cri-api-v1-PrivateAddressApi post the payload to get the session_id
- go the /Address endpoint make a request this should generate a 204 no-content result which signifies CRI is done and generated an authorizationCode
- use the authorizationCode to make a request against /callback endpoint on the headless stack for instance `ola-headless-stub-common`

Added `txma_stack_name` to the deploy.sh script in `test-resources` to help deploying a stack to host the endpoint for
`/callback` for example `ola-headless-stub-common` you can use the `deploy.sh` to deploy your copy if needed

The `client.ts` file was copied over [stub-oauth-client ](https://github.com/govuk-one-login/stub-oauth-client/blob/main/src/client.ts)from and called `crypto-service.ts` and used to generate the private_key_jwt [client authentication](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) i.e  https://github.com/govuk-one-login/stub-oauth-client/blob/main/test/create-private-key-jwt-request.test.ts#L6

needed for making a request against the public /token and point

image

### Why did it change

We want a headless, typescript core stub a /callback endpoint that takes an authorization code and returns the decrypted VC that was generated for the user.

- [OJ-2983](https://govukverify.atlassian.net/browse/OJ-2983)

### Other considerations

[OJ-2983]: https://govukverify.atlassian.net/browse/OJ-2983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test-resources/deploy.sh
```diff
@@ -27,4 +28,6 @@ sam deploy --stack-name "$stack_name" \
   cri:deployment-source=manual \
   cri:stack-type=localdev \
   --parameter-overrides \
-  Environment=localdev 
+  Environment=dev \
```
Caitlin-cooling, line 31: Is there a reason we changed this from `localdev`?

test-resources/package.json
```diff
@@ -16,7 +16,12 @@
     "dependencies": {
         "@aws-lambda-powertools/commons": "2.16.0",
         "@aws-lambda-powertools/logger": "2.16.0",
-        "esbuild": "0.25.1"
+        "@aws-lambda-powertools/parameters": "^2.16.0",
```
Caitlin-cooling, line 19: Please pin to a specific version

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,87 @@
...
-    async handler(): Promise<{ status: string; body: string }> {
-        return {
-            status: "200",
-            body: "Hello callback",
-        };
+    async handler(event: APIGatewayProxyEvent, _context: Context): Promise<APIGatewayProxyResult> {
+        try {
+            const authorizationCode = event.queryStringParameters?.authorizationCode;
+
+            if (!authorizationCode) {
```
Caitlin-cooling, line 17: Is this possible with the API GW config that we have?

test-resources/headless-core-stub/lambdas/callback/src/services/configuration-helper.ts
```diff
@@ -0,0 +1,34 @@
+import { getParametersByName } from "@aws-lambda-powertools/parameters/ssm";
+
+const commonParameterPrefix = process.env.AWS_STACK_NAME || "common-cri-api";
+const testResourcesParameterPrefix = process.env.TEST_RESOURCES_STACK_NAME || "test-resources";
+
+export class ConfigurationHelper {
+    public async getParameters(clientId: string) {
```
Caitlin-cooling, line 7: @mikebeeby has a similar thing in his PR [here](https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/414/files#diff-2683b3e55be29012b71a316aeed5ee2cd802269e879a1a7ce14b09633d3652db). Don't mind which approach we take but please can we align?

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,87 @@
...
+            return this.respondWith(500, err.message);
+        }
     }
+
+    private logApiError(endpoint: string, status: number, body: string) {
+        logger.error(`Request to ${endpoint} failed with status ${status} body: ${body}`);
+    }
+
+    private formatAudience = (audience: string) =>
+        audience.includes("review-") ? audience.replace("review-", "api.review-") : audience;
```
Caitlin-cooling, line 82: Why do we need to do this?

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,87 @@
...
+            if (!credential.ok) {
+                this.logApiError(credentialEndpoint, credential.status, credentialResponseBody);
+            }
+            logger.info("Successfully called /credential/issue endpoint");
+
+            return this.respondWith(credential.status, credentialResponseBody);
+        } catch (error: unknown) {
+            const err = error as Error;
+            logger.error(err.message);
+            return this.respondWith(500, err.message);
```
Caitlin-cooling, line 73: Similar comment about the error handling - the /start PR has a slightly different pattern so let's make sure they're aligned. 

Also are we sure we definitely need this `respondWith` abstraction? Looks like it just formats an object which is pretty simple to read on it's own I think?

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,87 @@
...
+            logger.info(`Calling ${credentialEndpoint}`);
+            const credential = await callbackService.issueCredential(credentialEndpoint, tokenBody.access_token);
+            const credentialResponseBody = await credential.text();
+
+            if (!credential.ok) {
+                this.logApiError(credentialEndpoint, credential.status, credentialResponseBody);
+            }
+            logger.info("Successfully called /credential/issue endpoint");
+
+            return this.respondWith(credential.status, credentialResponseBody);
```
Caitlin-cooling, line 69: There is an AC in the ticket about "returns a raw VC" rather than just the credential response, like the current stub does. Would that be significantly more work?

test-resources/headless-core-stub/lambdas/callback/src/services/crypto-service.ts
```diff
@@ -0,0 +1,65 @@
...
+    const response = await kmsClient.send(signCommand);
+    if (!response.Signature) {
+        throw new Error(`Failed to sign JWT with KMS key ${keyId}`);
+    }
+    jwtParts.signature = sigFormatter.derToJose(Buffer.from(response.Signature).toString("base64"), "ES256");
+    return jwtParts.header + "." + jwtParts.payload + "." + jwtParts.signature;
+};
+
+export const buildPrivateKeyJwtParams = async (params: PrivateJwtParams, headers?: JWTHeaderParameters) => {
+    const signedJwt = await signJwt(params.customClaims as JWTPayload, params, headers);
```
Caitlin-cooling, line 56: [nit] do we really need to pass in `params.customClaims` if we are passing in the whole `params` object next?

test-resources/infrastructure/template.yaml
```diff
@@ -88,6 +94,55 @@ Resources:
...
+            TableName: !Sub "{{resolve:ssm:/${CommonStackName}/SessionTableName}}"
+        - Statement:
+            - Effect: Allow
+              Action:
+                - ssm:GetParametersByPath
+                - ssm:GetParameters
+                - ssm:GetParameter
+              Resource:
+                - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${CommonStackName}/clients/ipv-core-stub-aws-headless/jwtAuthentication/*"
+                - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${CommonStackName}/SessionTableName"
```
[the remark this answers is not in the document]
jkunle: Yes, will remove SSM

test-resources/headless-core-stub/lambdas/callback/src/services/configuration-helper.ts
jkunle: I and @mikebeeby  agreed we will end up duplicating stuff and would come back to refactor

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
jkunle: Once the stub is live it would typically need `https://api.review-a.dev.account.gov.uk/token` or the credential one.  However, during test i.e integration-test for instance this would be public aws endpoint of the throw away stack.

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
jkunle: Was so happy to see something that looked like the VC, i forgot about it

test-resources/headless-core-stub/lambdas/callback/src/services/crypto-service.ts
```diff
@@ -0,0 +1,65 @@
...
+    } else if ("privateSigningKey" in params && params.privateSigningKey && isJWK(params.privateSigningKey)) {
+        const signingKey = await importJWK(params.privateSigningKey as JWK, "ES256");
+        signedJwt = await new SignJWT(jwtPayload).setProtectedHeader(jwtHeader).sign(signingKey);
+    } else {
+        throw new Error("No signing key provided!");
+    }
+    return signedJwt;
+};
+
+const signJwtViaKms = async (header: JWTHeaderParameters, payload: JWTPayload, keyId: string) => {
```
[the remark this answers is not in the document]
jkunle: It tries to do a lot not required in our current code, sorry it is not clear the intention. I may streamline this code but really just didn't touch it right apart from using it as is. I suppose it is reminder that we may want to use the `keyId` directly from KMS. I think @mikebeeby was also thinking the same. I didn't want to spend time on it for now so really we store the key in parameter store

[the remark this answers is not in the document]
Caitlin-cooling: Ok sure - so can we remove it for now?

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
Caitlin-cooling: Ok nice, yeah I would be in favor of us removing now that we are done debugging as API GW should give us a very similar errros

test-resources/headless-core-stub/lambdas/callback/src/services/configuration-helper.ts
Caitlin-cooling: Nice, is it worth catching up on that before we merge?

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,91 @@
...
+    async handler(event: APIGatewayProxyEvent, _context: Context): Promise<APIGatewayProxyResult> {
+        try {
+            const authorizationCode = event.queryStringParameters?.authorizationCode as string;
+            logger.info({ message: "Received authorizationCode", authorizationCode });
+
+            const sessionItem = await callbackService.getSessionByAuthorizationCode(
+                sessionTableName,
+                authorizationCode,
+            );
+            logger.info({ message: "Fetching session item...", ...sessionItem });
```
surajkumar, line 23: `Fetched`

surajkumar: Can we make sure the correct verb is used in other places too so we don't get confused on when the action happens - just gives clarity when we debug in the future.

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,91 @@
...
+            logger.info({ message: "Using Audience", audienceApi });
+
+            const tokenEndpoint = `${audienceApi}/token`;
+            logger.info({ message: "Calling token endpoint", tokenEndpoint, privateJwtParams });
+            const tokenResponse = await callbackService.getToken(tokenEndpoint, privateJwtParams);
+
+            if (!tokenResponse.ok) {
+                const tokenResponseBody = await tokenResponse.text();
+                const status = tokenResponse.status;
+                logger.error({ message: "Request to endpoint failed", tokenEndpoint, status, tokenResponseBody });
```
surajkumar, line 52: can we label the fields? e.g. 
`{ message: "Request to endpoint failed", endpoint: tokenEndpoint, statusCode: status, response: tokenResponseBody }`

surajkumar: Similarly for the other log messages. Unless they translate correctly in CloudWatch.

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,91 @@
...
+            );
+            logger.info({ message: "Fetching session item...", ...sessionItem });
+
+            const ssmParameters = await configurationHelper.getParameters(sessionItem.clientId);
+            const { privateSigningKey, ...filteredParams } = ssmParameters;
+
+            logger.info({ message: "Fetching SSM parameters", ...filteredParams });
+
+            const privateJwtKey = JSON.parse(privateSigningKey) as JWK;
+            const audience = ssmParameters["audience"];
```
surajkumar, line 31: This handler is pretty big, perhaps we should split it into logical methods e.g.

```
async function callTokenEndpoint(...) {
            const privateJwtKey = JSON.parse(privateSigningKey) as JWK;
            const audience = ssmParameters["audience"];

            logger.info({ message: "Generating private JWT parameters..." });
            const privateJwtParams = await generatePrivateJwtParams(
                sessionItem.clientId,
                authorizationCode,
                sessionItem.redirectUri,
                privateJwtKey,
                audience,
            );

            const audienceApi = this.formatAudience(audience);
            logger.info({ message: "Using Audience", audienceApi });

            const tokenEndpoint = `${audienceApi}/token`;
            logger.info({ message: "Calling token endpoint", tokenEndpoint, privateJwtParams });
            const tokenResponse = await callbackService.getToken(tokenEndpoint, privateJwtParams);
            logger.info({ message: "Successfully called /token endpoint" });
            const tokenBody = await tokenResponse.json();
          
           return {
              statusCode: tokenResponse.status
              body: tokenBody
           }
}
```

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -1,12 +1,91 @@
...
+                const tokenResponseBody = await tokenResponse.text();
+                const status = tokenResponse.status;
+                logger.error({ message: "Request to endpoint failed", tokenEndpoint, status, tokenResponseBody });
+                return { statusCode: tokenResponse.status, body: tokenResponseBody };
+            }
+
+            logger.info({ message: "Successfully called /token endpoint" });
+            const tokenBody = await tokenResponse.json();
+
+            const credentialEndpoint = `${audienceApi}/credential/issue`;
```
surajkumar, line 59: The credential endpoint stuff can also go in its own method

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
```diff
@@ -0,0 +1,53 @@
+import { DynamoDBClient, QueryCommand } from "@aws-sdk/client-dynamodb";
+import { SessionItem } from "./session-item";
+
+export class CallBackService {
+    constructor(private readonly dynamoDbClient = new DynamoDBClient({ region: process.env.REGION })) {}
+
+    public async getSessionByAuthorizationCode(sessionTable: string, code: string): Promise<SessionItem> {
+        const sessionItemQuery = await this.dynamoDbClient.send(
```
surajkumar, line 8: We'll need to add a re-try! It's querying a GSI

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        return {
+            sessionId: sessionItem?.sessionId?.S,
+            clientId: sessionItem?.clientId?.S,
+            authorizationCode: sessionItem?.authorizationCode?.S,
+            redirectUri: sessionItem?.redirectUri?.S,
+        } as SessionItem;
+    }
+
+    public async getToken(tokenUrl: string, body: string): Promise<Response> {
```
surajkumar, line 35: `callTokenEndpoint`

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
```diff
@@ -0,0 +1,53 @@
...
+        return await fetch(tokenUrl, {
+            method: "POST",
+            headers: {
+                "Content-Type": "application/x-www-form-urlencoded;charset=UTF-8",
+            },
+            body: body,
+        });
+    }
+
+    public async callIssueCredential(credentialUrl: string, accessToken: string): Promise<Response> {
```
surajkumar, line 45: `callIssueCredentialEndpoint`

surajkumar (changes requested):

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
jkunle: Hmmh, I thought you liked it this way

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
jkunle: Maybe not in this PR, will be future improvement but good to point out

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
surajkumar: Yeah to some extent, the flow being:

```
- do this
- then this
- then this
- more stuff
- done
```

as opposed to:

```
- do something
- then this

... 
do something:
  - do this
  - do something else

do something else:
  - do this... 
```

Basically, the logic going too far away from the handler. 

In this instance, we're keeping it within the handler but "hiding" the implementation as we just want to know what's happening in terms of logical flow.

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
surajkumar: Any reason we can just wrap this in a for-loop e.g. 

```
let item;
for(i .. 3) {
  // existing logic
 if (count != 0) {
    item = get(0);
    break;
   }
}
if(!item) throw Error();
return item;
```

I'm thinking we do this now because we're updating tests after this and the QAs might end up adding re-tries over there.

surajkumar: Similarly, other teams might be adopting this so we don't want known error scenarios being present if the fix the simple.

surajkumar: Similarly, other teams might be adopting this so we don't want known error scenarios being present if the fix the simple.

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
jkunle: We might not need this, it was useful during testing however. We change the audience of the our development stack to the public endpoint or localhost endpoint.

surajkumar (approved):

PR #428, opened 2025-04-02
LIME-1592 Added support for key rotation 

SarahHillGDS:
### What changed

- Introduced logic to decrypt JWE content using primary, secondary, and previous key aliases behind a feature toggle.

- Added a fallback mechanism to attempt decryption with the secondary and previous aliases if the primary alias fails during execution of the key rotation step function.

- Added unit test cases to verify decryption using each key alias, including fallback scenarios.

### Why did it change

The KMSRSADecrypter has been updated to support the use of KMS aliases for decryption. This change is required for the key rotation pilot in Passport CRI.

### Issue tracking

- [LIME-1592](https://govukverify.atlassian.net/browse/LIME-1592)

[LIME-1592]: https://govukverify.atlassian.net/browse/LIME-1592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ ext {
 		mockito                  : "4.3.1",
 		glassfish_version        : "3.0.3",
 		cri_common_lib           : "4.0.0",
+		webcompere_version       : "2.1.7",
 	]
 }
 
@@ -97,6 +98,7 @@ subprojects {
 		cucumberRuntime
 		cri_common_lib
 		pact_tests
+		webcompere
 	}
 
 	configurations.all {
@@ -144,7 +146,9 @@ subprojects {
 				"org.junit.jupiter:junit-jupiter-params:${dependencyVersions.junit}",
 				"org.mockito:mockito-junit-jupiter:${dependencyVersions.mockito}",
 				"org.mockito:mockito-inline:${dependencyVersions.mockito}",
-				"org.hamcrest:hamcrest:2.2"
+				"org.hamcrest:hamcrest:2.2",
+				"uk.org.webcompere:system-stubs-core:${dependencyVersions.webcompere_version}",
+				"uk.org.webcompere:system-stubs-jupiter:${dependencyVersions.webcompere_version}"
 
 		pact_tests "au.com.dius.pact.provider:junit5:4.5.8",
 				"au.com.dius.pact:provider:4.5.8",
--- a/deploy.sh
+++ b/deploy.sh
@@ -29,6 +29,6 @@ sam deploy --stack-name "$stack_name" \
   cri:deployment-source=manual \
   cri:stack-type=localdev \
   --parameter-overrides \
-  Environment=localdev \
+  Environment=dev \
   ${audit_event_name_prefix:+AuditEventNamePrefix=$audit_event_name_prefix} \
   ${cri_identifier:+CriIdentifier=$cri_identifier}
--- a/session/src/main/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypter.java
+++ b/session/src/main/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypter.java
@@ -9,6 +9,8 @@ import com.nimbusds.jose.crypto.impl.AlgorithmSupportMessage;
 import com.nimbusds.jose.crypto.impl.ContentCryptoProvider;
 import com.nimbusds.jose.jca.JWEJCAContext;
 import com.nimbusds.jose.util.Base64URL;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.services.kms.KmsClient;
 import software.amazon.awssdk.services.kms.model.DecryptRequest;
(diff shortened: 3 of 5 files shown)


build.gradle
```diff
@@ -158,6 +160,9 @@ subprojects {
 
 		nimbus "com.nimbusds:oauth2-oidc-sdk:${dependencyVersions.nimbusds_oauth_version}",
 				"com.nimbusds:nimbus-jose-jwt:${dependencyVersions.nimbusds_jwt_version}"
+
+		webcompere "uk.org.webcompere:system-stubs-core:${dependencyVersions.webcompere_version}",
```
surajkumar, line 164: I can't see where `webcompere` is being used in your PR?

deploy.sh
```diff
@@ -29,6 +29,6 @@ sam deploy --stack-name "$stack_name" \
   cri:deployment-source=manual \
   cri:stack-type=localdev \
   --parameter-overrides \
-  Environment=localdev \
+  Environment=dev \
```
surajkumar, line 32: Any reason for this change? The deploy.sh is for manual deployment and it's recommended to use localdev for most cases when running this script locally.

session/src/main/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypter.java
```diff
@@ -65,19 +77,79 @@ public byte[] decrypt(
...
-        DecryptResponse decryptResponse = this.kmsClient.decrypt(decryptRequest);
-        SecretKey cek = new SecretKeySpec(decryptResponse.plaintext().asByteArray(), "AES");
+        DecryptResponse decryptResponse = null;
+        if (keyRotationEnabled) {
+            LOGGER.info("Key rotation enabled. Attempting to decrypt with key aliases.");
+            // During a key rotation we might receive JWTs encrypted with either the old or new key.
+            boolean aliasDecryptionSuccessful = false;
+
+            try {
+                decryptResponse =
```
surajkumar, line 87: Perhaps it would be a good idea to move these `kmsClient.decrypt` + logs to their own methods. Just makes the code a little bit easier to understand. If you return a boolean on those new methods, you can then use that as the value for `aliasDecryptionSuccessful` via a short circuit i.e. 

```
aliasDecryptionSuccessful = decryptWithPrimaryKeyAlias() || decryptWithSecondaryKeyAlias() || decryptWithPreviousKeyAlias();
```

Without looking closely, I thought for a moment that it was all the same logic, hence why I think it might be a good refactor to do.

session/build.gradle
```diff
@@ -15,7 +15,8 @@ dependencies {
 			configurations.nimbus,
 			configurations.kms,
 			configurations.sqs,
-			configurations.jackson
+			configurations.jackson,
+			configurations.webcompere
```
surajkumar, line 19: Since webcompere is only used for your unit tests, it would be better to make this:
```
testImplementation configurations.webcompere
```
Otherwise, webcompere might get bundled with the Lambda code that gets uploaded to AWS.

session/src/main/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypter.java
```diff
@@ -65,19 +77,79 @@ public byte[] decrypt(
...
+                try {
+                    decryptResponse =
+                            kmsClient.decrypt(
+                                    buildDecryptRequest(
+                                            SESSION_DECRYPTION_KEY_PREVIOUS_ALIAS, encryptedKey));
+                    LOGGER.info(
+                            "Decryption successful with previous key alias: {}",
+                            SESSION_DECRYPTION_KEY_PREVIOUS_ALIAS);
+                } catch (Exception e) {
+                    LOGGER.error("Failed to decrypt with all available key aliases.");
```
surajkumar, line 126: Would be good to include the error as part of your log statements, without it, support/developers are left to guess what the root cause might be, e.g. non-existent key? permissions? something else?

surajkumar (changes requested):

build.gradle
surajkumar: Ignore :)

deploy.sh
smsgds: deploy.sh is used by Lime to deploy into our aws dev account?

Which then would fail as Environment=localdev is invalid?

session/src/main/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypter.java
SarahHillGDS: Need to return the decryptResponse which made boolean methods challenging. Refactored kmsClient.decrypt into its own method that loops through the aliases and shows errors in logs.

session/build.gradle
SarahHillGDS: Moved to test configuration

session/src/main/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypter.java
SarahHillGDS: Added error logs to decryptWithKeyAliases

smsgds (changes requested):
Just needs commits squash to a single one.

session/src/test/java/uk/gov/di/ipv/cri/common/api/service/KMSRSADecrypterTest.java
```diff
@@ -60,38 +75,201 @@ void shouldDecrypt() throws ParseException, JOSEException {
                                                 .decode(
                                                         "ngoABokVaj3BYY8FfaPef4nzV9dr+ziueibf2hofYDQ=")))
                         .build();
+        System.out.println(decryptResponse.toString());
```
gds-jbentham, line 78: Can we remove this system print line

deploy.sh
surajkumar: Ah interesting, thought it was in that list. I'll see if we can add localdev to. Good catch

smsgds (approved):

surajkumar (approved):

ElliotMurphyGDS (approved):

PR #442, opened 2025-04-09
OJ-3132 refactor to implement client_id and state query params to /callback

Caitlin-cooling:
### What changed

Refactor to implement client_id and state query params to /callback

### Why did it change

- To closer resemble what core is actually doing
- To remove the need for the session table

### Screenshots

image

### Issue tracking

- [OJ-3132](https://govukverify.atlassian.net/browse/OJ-3132)


test-resources/infrastructure/public-api.yaml
```diff
@@ -14,6 +14,18 @@ paths:
...
             type: string
+        - name: client_id
+          in: query
+          required: false
+          description: The OAuth client identifier
+          schema:
+            type: string
+        - name: state
+          in: query
+          required: false
```
jkunle, line 25: same here

test-resources/infrastructure/public-api.yaml
```diff
@@ -14,6 +14,18 @@ paths:
           description: authorizationCode generated by CRI
           schema:
             type: string
+        - name: client_id
+          in: query
+          required: false
```
jkunle, line 19: If core behaves like this, why isn't this `required: true`

test-resources/headless-core-stub/lambdas/callback/tests/callback-handler.test.ts
```diff
@@ -71,18 +64,92 @@ describe("callback-handler", () => {
...
-            body: "No session item found for provided authorizationCode",
+            statusCode: 200,
+            headers: {
+                "Content-Type": "text/plain",
+            },
+            body: "vc.jwt.credential",
+        });
+    });
+
+    it("uses state from request params if provided", async () => {
```
jkunle, line 106: nice, how about when this state is not provided? would be very nice to have a test for that

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
```diff
@@ -1,45 +1,13 @@
...
-import { SessionItem } from "./session-item";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
 import { APIGatewayProxyResult } from "aws-lambda";
 import { Logger } from "@aws-lambda-powertools/logger";
-import { withRetry } from "../../../../utils/src/retrier/retry";
 
 export class CallBackService {
     constructor(
         private readonly logger: Logger,
         private readonly dynamoDbClient = new DynamoDBClient({ region: process.env.REGION }),
```
jkunle, line 8: and this too

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
```diff
@@ -1,45 +1,13 @@
-import { DynamoDBClient, QueryCommand } from "@aws-sdk/client-dynamodb";
-import { SessionItem } from "./session-item";
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
```
jkunle, line 1: yay!, this could go

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
```diff
@@ -38,16 +53,19 @@ export class CallbackLambdaHandler implements LambdaInterface {
...
     }
 
-    private async generatePrivateJwtParams(session: SessionItem, code: string, ssmParameters: Record<string, string>) {
-        logger.info({ message: "Generating private JWT parameters" });
+    private extractFromState(state: string) {
+        let audience;
+        let redirectUri;
+
+        if (state) {
+            const statePayload = JSON.parse(base64Decode(state));
```
mikebeeby, line 61: If state is not base64 encoded Json, this will throw a error, resulting in a 500 error. Should really be a 400

test-resources/headless-core-stub/lambdas/callback/src/services/callback-service.ts
Caitlin-cooling: Nice spot! Done

test-resources/infrastructure/public-api.yaml
jkunle: I think it is 
https://github.com/govuk-one-login/ipv-core-back/blob/main/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java#L154

then `input body` of the core-callback Request is https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CriCallbackRequest.java#L19

We are interested in:
    private String authorizationCode;
    private String credentialIssuerId;
    private String redirectUri;

I think `credentialIssuerId` is an identifier that can represent the audience, given that this is an `internal api` i think we can just send the audience and the redirect url as query params instead of using state

jkunle: So ipv-core frontend corresponds as https://github.com/govuk-one-login/ipv-core-front/blob/main/src/app/credential-issuer/middleware.ts#L32-L46

test-resources/headless-core-stub/lambdas/callback/src/callback-handler.ts
Caitlin-cooling: Yeah good shout - added some error handling

test-resources/headless-core-stub/lambdas/callback/tests/callback-handler.test.ts
Caitlin-cooling: That would be covered in the `returns 200 when entire flow is successful` test

test-resources/infrastructure/public-api.yaml
Caitlin-cooling: I'm wondering if it's this middleware from Core that we should be looking at actually https://github.com/govuk-one-login/ipv-core-front/blob/main/src/app/oauth2/middleware.ts

jkunle (approved):

PR #447, opened 2025-04-23
feat: set environment variables to use key rotation

surajkumar:
### What changed
bump cri-lib to 5.0.0
set `ENV_VAR_FEATURE_CONSUME_PUBLIC_JWK` environment variable for token and session lambda
set `PUBLIC_JWKS_ENDPOINT ` environment variable for token and session lambda

### Why did it change
So that we can use cores public JWK endpoint

## Note
This is only going to be abled for address-cri and if you want to add other CRIs, you have to append to `KeyRotationEnabled` and `JwkEndpoint` mapping with the cri identifier. 

### Issue tracking
- [OJ-2972](https://govukverify.atlassian.net/browse/OJ-2972)

[OJ-2972]: https://govukverify.atlassian.net/browse/OJ-2972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -169,6 +169,94 @@ Globals:
 
 Mappings:
 
+  KeyRotationEnabled:
+    di-ipv-cri-address-api:
+      dev: true
+      build: true
+      staging: true
+      integration: false
+      production: false
+    di-ipv-cri-fraud-api:
+      dev: false
+      build: false
+      staging: false
+      integration: false
+      production: false
+    di-ipv-cri-kbv-api:
+      dev: false
+      build: false
+      staging: false
+      integration: false
+      production: false
+    di-ipv-cri-dl-api:
+      dev: false
+      build: false
+      staging: false
+      integration: false
+      production: false
+    di-ipv-cri-passport-api:
+      dev: false
+      build: false
+      staging: false
+      integration: false
+      production: false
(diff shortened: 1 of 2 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -659,7 +747,8 @@ Resources:
           POWERTOOLS_SERVICE_NAME: !Sub "${CriIdentifier}-session"
           SQS_AUDIT_EVENT_QUEUE_URL:
             Fn::ImportValue: !Sub ${TxmaStackName}-AuditEventQueueUrl
-          ENV_VAR_FEATURE_FLAG_KEY_ROTATION: !FindInMap [ KeyRotationMapping, !Ref CriIdentifier, !Ref Environment]
```
Caitlin-cooling, line 662: We don't want to remove this one I don't think? Lime added it for the implementation of our own jwks endpoints piece - this is the OJ ticket https://govukverify.atlassian.net/browse/OJ-3143

surajkumar: Good catch - reverted!

Caitlin-cooling:
Pls share with Lime!

jkunle (approved):

PR #450, opened 2025-04-25
OJ-2978

jkunle:
### Issue tracking

### Other considerations


infrastructure/lambda/template.yaml
```diff
@@ -737,6 +737,17 @@ Resources:
       CodeUri: ../../lambdas
       Handler: session-handler.lambdaHandler
       Runtime: nodejs22.x
+      VpcConfig:
```
Caitlin-cooling, line 740: Why is this needed?

test-resources/headless-core-stub/lambdas/mock-jwks/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "headless-core-stub-mock-jwk",
```
Caitlin-cooling, line 2: [nit] `headless-core-stub-mock-jwks`

test-resources/headless-core-stub/utils/src/services/client-configuration.ts
```diff
@@ -1,14 +1,15 @@
...
         const parameters = [
-            `/${commonParameterPrefix}/clients/${clientId}/jwtAuthentication/audience`,
-            `/${commonParameterPrefix}/clients/${clientId}/jwtAuthentication/issuer`,
-            `/${commonParameterPrefix}/clients/${clientId}/jwtAuthentication/redirectUri`,
-            `/${testResourcesParameterPrefix}/ipv-core-stub-aws-headless/privateSigningKey`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/audience`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/issuer`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/redirectUri`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/publicSigningJwkBase64`,
+            "/test-resources/ipv-core-stub-aws-headless/privateSigningKey",
```
Caitlin-cooling, line 12: Why are we hardcoding the `testResourcesParameterPrefix` now?

infrastructure/lambda/template.yaml
jkunle: no longer needed

test-resources/headless-core-stub/utils/src/services/client-configuration.ts
jkunle: It's always the [same value](https://github.com/govuk-one-login/ipv-cri-pipeline-deployment/blob/main/config/shared/output.tf#L2) this make it easier to work with, when you have your own test resources stack

test-resources/infrastructure/public-api.yaml
```diff
@@ -4,6 +4,54 @@ info:
...
+            method.response.header.X-Content-Type-Options: "'nosniff'"
+            method.response.header.X-Frame-Options: "'DENY'"
+            method.response.header.Content-Type: "'application/json'"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/JsonWebKeySet"
+        "404":
+          description: Not Found
+      x-amazon-apigateway-request-validator: "Validate both"
```
Caitlin-cooling, line 44: What are we validating if this request doesn't accept a body or request params?

test-resources/headless-core-stub/utils/src/services/client-configuration.ts
Caitlin-cooling: Hmm I think in this context it would be very common to deploy you're own test resources stack and although it is the same value now we can't assume that will always be the case.

Is this being a param causing some sort of issue that we are trying to fix/

test-resources/headless-core-stub/lambdas/mock-jwks/src/services/cache-jwk.ts
```diff
@@ -0,0 +1,33 @@
...
+import { CORE_STUB_SIGNING_PUBLIC_JWK } from "../constants";
+
+let cachedKeys: { jwks: JSONWebKeySet } | null = null;
+
+export const generateJWKS = async (clientId: string): Promise<{ jwks: JSONWebKeySet; privateKeys?: JWK[] }> => {
+    if (cachedKeys) {
+        return cachedKeys;
+    }
+
+    const ssmParameters = await ClientConfiguration.getConfig(clientId);
```
Caitlin-cooling, line 15: Ah so I guess this would be a good use case for a getParameter function that just gets one param rather than all of them?

test-resources/headless-core-stub/lambdas/mock-jwks/src/services/cache-jwk.ts
```diff
@@ -0,0 +1,33 @@
...
+    const currentPublicKey = CORE_STUB_SIGNING_PUBLIC_JWK as JWK;
+    const currentPrivateKey = JSON.parse(ssmParameters.privateSigningKey) as JWK;
```
Caitlin-cooling, line 17: Sorry I'm really not following what we are doing here with private keys. We seem to add them to an object and then not use them? What am I missing?

test-resources/headless-core-stub/lambdas/mock-jwks/src/mock-jwk-handler.ts
```diff
@@ -0,0 +1,36 @@
...
+import { handleErrorResponse } from "../../../utils/src/errors/error-response";
+import { generateJWKS } from "./services/cache-jwk";
+
+const logger = new Logger({ serviceName: "Mock Jwk Handler" });
+
+export class MockJwkHandler implements LambdaInterface {
+    async handler(event: APIGatewayProxyEvent, _context: Context): Promise<APIGatewayProxyResult> {
+        try {
+            logger.info({ message: "Retrieving JWKS", path: event.path });
+            const { jwks } = await generateJWKS("ipv-core-stub-aws-headless");
```
Caitlin-cooling, line 13: Should we make this magic string a constant?

Caitlin-cooling: There actually is already `DEFAULT_CLIENT_ID`

test-resources/headless-core-stub/lambdas/start/tests/start-handler.test.ts
```diff
@@ -45,15 +42,15 @@ describe("start-handler", () => {
     });
 
     it("returns 200 when body is empty", async () => {
-        expect(true);
```
Caitlin-cooling, line 48: Oh nice!

test-resources/infrastructure/public-api.yaml
```diff
@@ -4,6 +4,55 @@ info:
   version: "1.0"
 
 paths:
+  /.well-known/jwks.json:
```
Caitlin-cooling, line 7: Please could we have a happy path integration test for this new endpoint?

test-resources/integration-tests/start/start-happy.test.ts
```diff
@@ -1,51 +1,54 @@
 import { stackOutputs } from "../helpers/cloudformation";
-import { signedFetch } from "../helpers/fetch";
 import { kmsClient } from "../helpers/kms";
 import { JweDecrypter } from "../helpers/jwe-decrypter";
 import { JwtVerifierFactory, ClaimNames } from "../helpers/jwt-verifier";
 import { getParametersValues } from "../../headless-core-stub/utils/src/parameter/get-parameters";
-
+import { base64Encode } from "../../headless-core-stub/utils/src/base64/index";
+import { Logger } from "@aws-lambda-powertools/logger";
+import { signedFetch } from "../helpers/fetch";
```
Caitlin-cooling, line 8: [nit] it's generally good practise for imports to be in alpahateical order because it's easier to read

test-resources/integration-tests/start/start-happy.test.ts
```diff
@@ -1,51 +1,54 @@
...
+    const jwtVerifierFactory = new JwtVerifierFactory(new Logger());
     const clientId = "ipv-core-stub-aws-headless";
     const aud = "https://test-aud";
-    const iss = "https://test-issuer";
+    let iss: string;
 
     beforeAll(async () => {
         const { TestHarnessExecuteUrl, CommonStackName } = await stackOutputs(process.env.STACK_NAME);
         testHarnessExecuteUrl = TestHarnessExecuteUrl;
+        iss = TestHarnessExecuteUrl.replace(/\/+$/, "");
```
Caitlin-cooling, line 22: Why do we have to do this?

test-resources/integration-tests/start/start-happy.test.ts
```diff
@@ -1,51 +1,54 @@
...
 
         jweDecrypter = new JweDecrypter(kmsClient, () => decryptionKeyId);
     });
 
     it("returns 200 with a valid JWT for a valid request", async () => {
-        // Decodes to {"aud":"https://test-aud","redirect_uri":"https://test-resources.review-hc.dev.account.gov.uk/callback"}
-        const defaultState =
-            "eyJhdWQiOiJodHRwczovL3Rlc3QtYXVkIiwicmVkaXJlY3RfdXJpIjoiaHR0cHM6Ly90ZXN0LXJlc291cmNlcy5yZXZpZXctaGMuZGV2LmFjY291bnQuZ292LnVrL2NhbGxiYWNrIn0="; // pragma: allowlist secret
-        const data = await signedFetch(`${testHarnessExecuteUrl}start`, {
+        const defaultState = base64Encode(
```
Caitlin-cooling, line 34: Hmm I hard coded this becuase I didn't want our tests to rely on our helpers - if they break then tests for code that don't use those helpers break.

Keen to hear why you think doing it like this is better?

test-resources/integration-tests/helpers/jwt-verifier.ts
```diff
@@ -21,44 +25,55 @@ export enum ClaimNames {
...
     ): Promise<JWTVerifyResult | null> {
-        try {
-            const signingPublicJwkBase64 = this.jwtVerifierConfig.publicSigningJwk;
-            const signingAlgorithm = this.jwtVerifierConfig.jwtSigningAlgorithm;
-            const signingPublicJwk = JSON.parse(Buffer.from(signingPublicJwkBase64, "base64").toString("utf8"));
-            const publicKey = await importJWK(signingPublicJwk, signingPublicJwk?.alg || signingAlgorithm);
-
-            const jwtVerifyOptions = this.createJwtVerifyOptions(signingAlgorithm, expectedClaimValues);
-            const verifyResult = await jwtVerify(encodedJwt, publicKey, jwtVerifyOptions);
+        if (!mandatoryClaims || mandatoryClaims?.size === 0) throw new Error("No mandatory claims provided");
```
Caitlin-cooling, line 38: I didn't put error handling in this because I expected it to be caught by teest assertions. Keen to understand why you disagree?

test-resources/integration-tests/helpers/jwt-verifier.ts
```diff
@@ -21,44 +25,55 @@ export enum ClaimNames {
...
+    }
 
-    private createJwtVerifyOptions(
-        signingAlgorithm: string,
-        expectedClaimValues: Map<string, string>,
-    ): JWTVerifyOptions {
+    private getPublicKeyUsingJwksUri(issuer?: string) {
+        this.logger.info({ message: "getPublicKeyUsingJwksUri:- using issuer", issuer });
+
+        this.logger.info("Retrieving publicSigningKey from JwkEndpoint");
```
Caitlin-cooling, line 67: Can these two logs be one message do you think?

test-resources/headless-core-stub/lambdas/mock-jwks/src/services/cache-jwk.ts
jkunle: I use them in the test, as it is a `keypair`, to show `publickey.kid === privatekey.kid`

test-resources/integration-tests/start/start-happy.test.ts
jkunle: It to remove `https://test-resources.review-hc.dev.account.gov.uk/`

test-resources/integration-tests/start/start-happy.test.ts
jkunle: When I ran this with my test-resources stack it broke, 

`redirect_uri: `https://${TEST_RESOURCES_STACK_NAME}.review-hc.dev.account.gov.uk/callback`,

Then it passed on my stack, but when I pushed it broke, so I change it back, but no longer hard-coded b'cos it not obvious before that it has to be `https://test-resources.review-hc.dev.account.gov.uk/callback`

I believe https://github.com/govuk-one-login/ipv-cri-common-lambdas/blob/main/infrastructure/lambda/template.yaml#L1779 maybe need to move into the `test-resources template` am torn on that, anyway this is why it's no longer hard-coded

test-resources/integration-tests/helpers/jwt-verifier.ts
jkunle: No, I copied over the original jwtVerifier I played around with the `getPublicKeyUsingJwksUri` in my stack first

test-resources/infrastructure/public-api.yaml
jkunle: I didn't think we needed a separate test for it that why I reworked `JwtVerifier` -> `getPublicKeyUsingJwksUri` which calls it.

test-resources/headless-core-stub/utils/src/services/client-configuration.ts
```diff
@@ -1,14 +1,16 @@
...
         const parameters = [
-            `/${commonParameterPrefix}/clients/${clientId}/jwtAuthentication/audience`,
-            `/${commonParameterPrefix}/clients/${clientId}/jwtAuthentication/issuer`,
-            `/${commonParameterPrefix}/clients/${clientId}/jwtAuthentication/redirectUri`,
-            `/${testResourcesParameterPrefix}/ipv-core-stub-aws-headless/privateSigningKey`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/audience`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/issuer`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/redirectUri`,
+            `/${commonStackName}/clients/${clientId}/jwtAuthentication/publicSigningJwkBase64`,
+            STUB_PRIVATE_SIGNING_KEY_PARAMETER_PATH,
```
Caitlin-cooling, line 13: Why are we not following the same pattern for this as the other SSM params?

test-resources/integration-tests/start/start-happy.test.ts
Caitlin-cooling: Hmm what? Looks like it removes the trailing slash, is that what you mean? 
image

test-resources/headless-core-stub/lambdas/mock-jwks/src/services/cache-jwk.ts
Caitlin-cooling: Which test? I wonder if there's a way around it all all? Doesn't feel quite right to give a lambda knowledge of a private key just for testing

test-resources/headless-core-stub/utils/src/services/client-configuration.ts
jkunle: It an absolute path parameter just like https://github.com/govuk-one-login/ipv-cri-pipeline-deployment/blob/main/config/check-hmrc/build/ssm/main.tf#L22-L40

test-resources/integration-tests/start/start-happy.test.ts
Caitlin-cooling: Ok so if we have to do manipulation is stack outputs the right thing to be using for this? Should we just use the SSM param?

test-resources/headless-core-stub/utils/src/services/client-configuration.ts
Caitlin-cooling: Sorry maybe I wasn't clear - what is the reason are we not following the pattern this is being used in this file for other SSM params?

jkunle: You mean  this pattern`/${testResourcesParameterPrefix}/${clientId}/privateSigningKey` 
As mention the parameter is really an absolute one all stubs stacks would be using it. Include one the you would create you had you test-resources stack for instance, It does change stack by stack

jkunle: The parameter could have been named `/ipv-core-stub-aws-headless/privateSigningKey` it is currently used in `dev, build` by ours stubs and cross all other flavoured existing stub environment stacks

Caitlin-cooling: Right ok you're saying that the SSM param isn't deployed with the test-harness stack itself so using the name of the test-harness stack isn't helpful? 

FIne, happy for us to hard code "test-resources" then but let's do it in this file rather than as a constant since it's not reused and we need the `clientId` override since that is possible to override everywhere else in the stub.

test-resources/integration-tests/start/start-happy.test.ts
Caitlin-cooling: I think you fixed this but it's showing out of order again?

Caitlin-cooling (dismissed):

test-resources/headless-core-stub/lambdas/mock-jwks/src/services/cache-jwk.ts
```diff
@@ -0,0 +1,30 @@
...
+export const generateJWKS = async (): Promise<{ jwks: JSONWebKeySet }> => {
+    if (cachedKeys) {
+        return cachedKeys;
+    }
+
+    const currentPublicKey = CORE_STUB_SIGNING_PUBLIC_JWK as JWK;
+
+    const [key1, key2] = await Promise.all([
+        getJwkKeyPair({ currentPublicKey } as GetJwkKeyPairOptions),
+        getJwkKeyPair({ kid: getHashedKid(randomUUID()) }),
```
surajkumar, line 18: Q: If we are using a randomUUID() here for the kid, how will this work for the integration-tests? Or does the start endpoint use this? Just trying to understand the flow here in my head as in my ticket we will be doing: 

```
signedJWT.getHeader().getKeyID()
```
and using the value received from here to find the correct entry your JWKS endpoint.

jkunle: `getJwkKeyPair({ kid: getHashedKid(randomUUID()) })` belongs to the new public signing key, we use the first key which is the first position and is the one we currently have a private key in SSM

Caitlin-cooling: Yeah the one with the random KID will never actually be used to sign a JWT so the session handler so should need to use it - it's just to replicate the conditions of a rotation where we have two keys returned (one used, and one unused)

Caitlin-cooling (approved):

PR #465, opened 2025-05-06
OJ-3075: Add jwt_verification_failed metric

Caitlin-cooling:
### What changed

Add `jwt_verification_failed` metric

### Why did it change

So that we can create an alarm to alert us to an increase in issues decrypting the JWTs as this may indicate an issue with core's jwks endpoint, or an issue with our logic that uses it, which would be considered a P1/P2.

I chose to add this in common lambdas rather than cri lib for consistency with other metrics and to act as a bit of a catch all. SessionValidationException, ClientConfigurationException and AccessTokenValidationException errors could be caused by more than just jwks endpoint related failures, but in general if we are consistently unable to verify requests that are coming into the CRI from core, this is something we want to know about ASAP.

Alarm added to address in https://github.com/govuk-one-login/ipv-cri-address-api/pull/1275

### Screenshots

Screenshot 2025-05-07 at 2 32 08 pm

### Issue tracking

- [OJ-3075](https://govukverify.atlassian.net/browse/OJ-3075)

[OJ-3075]: https://govukverify.atlassian.net/browse/OJ-3075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/session/src/test/java/uk/gov/di/ipv/cri/common/api/handler/SessionHandlerTest.java
+++ b/session/src/test/java/uk/gov/di/ipv/cri/common/api/handler/SessionHandlerTest.java
@@ -40,7 +40,12 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.isNull;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 import static uk.gov.di.ipv.cri.common.api.handler.SessionHandler.REDIRECT_URI;
 import static uk.gov.di.ipv.cri.common.api.handler.SessionHandler.STATE;
 


session/src/main/java/uk/gov/di/ipv/cri/common/api/handler/SessionHandler.java
```diff
@@ -141,12 +142,14 @@ public APIGatewayProxyResponseEvent handleRequest(
...
 
             eventProbe.log(ERROR, e).counterMetric(EVENT_SESSION_CREATED, 0d);
+            eventProbe.counterMetric(JWT_VERIFICATION_FAILED);
 
             return ApiGatewayResponseGenerator.proxyJsonResponse(
                     HttpStatusCode.BAD_REQUEST, ErrorResponse.SESSION_VALIDATION_ERROR);
         } catch (ClientConfigurationException | SqsException e) {
 
             eventProbe.log(ERROR, e).counterMetric(EVENT_SESSION_CREATED, 0d);
+            eventProbe.counterMetric(JWT_VERIFICATION_FAILED);
```
mikebeeby, line 152: Hmm I understand why you have done it like this, but it does come at a cost of reduced readability. From a quick glance now it would read that if any of these exceptions are thrown it's because of JWT Verification failing, even though that could have passed. 

It could also act as a red herring to support if the alarm for this metric is triggered but it's actually an SqsException for example.

Caitlin-cooling: Yeah I see your point - I guess what we are catching here is any JWT verification errors, not just errors when trying to verify the signing key.

My feeling is that the other causes (such as unexpected signing algorithm, or a bad JWT claimset) are also things that we would want to alarm on, even if they aren't directly related to this piece of work, because ultimately there is an issue with the JWTs that core is issuing.

I agree the `SqsException` doesn't make sense in this context if I move that into it's own catch, does that help? Or are you suggesting moving these in the [cri lib JWTVerifier](https://github.com/govuk-one-login/ipv-cri-lib/blob/ee04de8a0b2fcfb92294892eb6af68f37c547d70/src/main/java/uk/gov/di/ipv/cri/common/library/service/JWTVerifier.java)?

mikebeeby: Yeah split the SqsException, or wrap the metric in an `instanceof ClientConfigurationException`. I think that will be fine. I've had a little look/think and it could end up being a bit of a rabbit hole refactoring to make it more precise.

accesstoken/src/main/java/uk/gov/di/ipv/cri/common/api/handler/AccessTokenHandler.java
```diff
@@ -79,6 +80,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                     HttpStatusCode.OK, accessTokenResponse.toJSONObject());
         } catch (AccessTokenValidationException e) {
             eventProbe.log(Level.ERROR, e).counterMetric(METRIC_NAME_ACCESS_TOKEN, 0d);
+            eventProbe.counterMetric(JWT_VERIFICATION_FAILED);
```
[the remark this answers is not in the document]
Caitlin-cooling: Done!

session/src/main/java/uk/gov/di/ipv/cri/common/api/handler/SessionHandler.java
Caitlin-cooling: Cool - done

session/src/test/java/uk/gov/di/ipv/cri/common/api/handler/SessionHandlerTest.java
```diff
@@ -40,16 +40,14 @@
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
jkunle, line 43: Please could we use explicit imports rather than wild-cards

Caitlin-cooling: Done

mikebeeby (approved):

PR #466, opened 2025-05-07
LIME-1447 - disabling provisioned concurrency (int/prod)

ElliotMurphyGDS:
### What changed
This PR is to disable provisioned concurrency for frauds common-lambdas, there will be a fast follow PR to enable snapstart - https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/467


smsgds (approved):

Caitlin-cooling (approved):
Approving as the change looks sensible and Seamus has approved from a Lime side

PR #467, opened 2025-05-07
LIME-1447 - enabling snapstart (int/prod)

ElliotMurphyGDS:
### What changed
This PR is to enabled Snapstart for Frauds common lambdas in integration and production and should be merged as a fast follow to this PR which disabled provisioned concurrency - https://github.com/govuk-one-login/ipv-cri-common-lambdas/pull/466


smsgds (approved):

Caitlin-cooling (approved):
Approving as the change looks sensible and Seamus has approved from a Lime side

PR #472, opened 2025-05-12
OJ-3078: Change the subnet of IsDevPlatformDeploy lambdas to protected

mikebeeby:
### What changed

Added an additional `If` to `IsDevPlatformDeploy` for selecting which subnet to use. `IsKeyRotationEnabled` also being true will set the subnet to `protected`.

### Why did it change

Session and Access Token lambdas now need to make a call to core's JWKS endpoint or a stubbed implementation (test-resources). They cannot do this from the private subnet when they are using the newer VPC set up, so need moving to the protected subnet. 

Initially limiting testing with `check-hmrc` dev

### Issue tracking

- [OJ-3078](https://govukverify.atlassian.net/browse/OJ-3078)

[OJ-3078]: https://govukverify.atlassian.net/browse/OJ-3078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -78,6 +78,9 @@ Conditions:
   IsDevPlatformDeploy: !Equals
     - !FindInMap [CriVpcMapping, !Ref CriIdentifier, !Ref Environment]
     - "di-devplatform-deploy"
+  IsKeyRotationEnabled: !Equals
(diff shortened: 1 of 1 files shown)


mikebeeby:
Closed in favour of new ticket: https://govukverify.atlassian.net/browse/OJ-3181

PR #473, opened 2025-05-14
OJ-3080: Add jwt_verification_failed metric to TS session and token

mikebeeby:
### What changed

Added `jwt_verification_failed` metric to TS session and token

Note: There seems to be an issue with metrics not being published when errors are thrown by the `validateJwtMiddleware` and caught by the `errorMiddleware`. I have added `metrics.publishStoredMetrics();` in the errorMiddleware to make sure the new metric is published. This can/should be looked at in more detail if needed in the future.

### Why did it change

Now that we are consuming Cores public signing keys from an endpoint rather than SSM there is more risk that something will go wrong when verifying requests from Core

### Issue tracking

- [OJ-3080](https://govukverify.atlassian.net/browse/OJ-3080)

### Screenshot

image

[OJ-3080]: https://govukverify.atlassian.net/browse/OJ-3080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/src/middlewares/error/error-middleware.ts
+++ b/lambdas/src/middlewares/error/error-middleware.ts
@@ -1,8 +1,10 @@
 import { Logger } from "@aws-lambda-powertools/logger";
 import { Metrics, MetricUnits } from "@aws-lambda-powertools/metrics";
 import { MiddlewareObj, Request } from "@middy/core";
-import { errorPayload } from "../../common/utils/errors";
+import { errorPayload, SessionValidationError } from "../../common/utils/errors";
 
+const SESSION_CREATED_METRIC = "session_created";
+const JWT_VERIFICATION_FAILED_METRIC = "jwt_verification_failed";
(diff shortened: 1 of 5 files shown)


lambdas/src/middlewares/error/error-middleware.ts
```diff
@@ -16,6 +17,10 @@ const errorMiddleware = (
         if (request.response !== undefined) return;
 
         metrics.addMetric(options.metric_name, MetricUnits.Count, 0);
+        if (request.error instanceof SessionValidationError && options.metric_name === "session_created") {
```
Caitlin-cooling, line 20: Shall we use the `SESSION_CREATED_METRIC`  constant?

Caitlin-cooling (dismissed):

PR #474, opened 2025-05-16
OJ-2969: uses public encryption key from the CRI's ./well-known/jwks.json endpoint

jkunle:
## Proposed changes

Added functionality to call the hosted .well-known/jwks.json of the CRI.
The headless core stub using the jwk to encrypt the JAR

### What changed

The Key-rotation step function has been run and a bootstrap key is available for use, which can be used to encrypt the payload by the headless core stub. The CRI should be able to decrypt this request when it comes throught the session handler

### Why did it change

A valid ./well-known/jwks.json exist at https://api.review-a.dev.account.gov.uk/.well-known/jwks.json

image

### Issue tracking

- [OJ-2969](https://govukverify.atlassian.net/browse/OJ-2969)

### Other considerations

[OJ-2969]: https://govukverify.atlassian.net/browse/OJ-2969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
+++ b/test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
@@ -1,16 +1,63 @@
 import { GetPublicKeyCommand, KMSClient } from "@aws-sdk/client-kms";
-import { CompactEncrypt, importSPKI, KeyLike } from "jose";
+import { CompactEncrypt, importJWK, importSPKI, JSONWebKeySet, JWK, KeyLike } from "jose";
 import { HeadlessCoreStubError } from "../../../../utils/src/errors/headless-core-stub-error";
+import { formatAudience } from "../../../../utils/src/audience-formatter";
+import { Logger } from "@aws-lambda-powertools/logger";
 
 const kmsClient = new KMSClient({ region: "eu-west-2" });
+export const logger = new Logger();
 
 let cachedPublicKey: KeyLike | undefined;
 
-export const getPublicEncryptionKey = async () => {
+export const getPublicEncryptionKey = async (audience: string) => {
     if (cachedPublicKey) {
         return cachedPublicKey;
     }
 
+    if (process.env.KEY_ROTATION_FEATURE_FLAG_ENABLED === "true") {
+        return await getPublicEncryptionKeyJwksUri(audience);
+    }
+
+    if (!cachedPublicKey) {
+        logger.info({ message: "using KMS to retrieve encryption key" });
+        await getPublicEncryptionKeyFromKms();
+    }
+
+    return cachedPublicKey;
+};
+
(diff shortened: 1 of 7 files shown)


test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
```diff
@@ -1,16 +1,65 @@
...
 let cachedPublicKey: KeyLike | undefined;
 
-export const getPublicEncryptionKey = async () => {
+export const getPublicEncryptionKey = async (audience: string) => {
     if (cachedPublicKey) {
         return cachedPublicKey;
     }
 
+    try {
+        await getPublicEncryptionKeyFromJwsUri(audience);
```
Caitlin-cooling, line 18: typo? `getPublicEncryptionKeyFromJwksUri`

test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
```diff
@@ -1,16 +1,65 @@
...
+    } catch (err) {
+        logger.warn({ message: "Failed to retrieve key from JWS URI, falling back to KMS", err });
```
Caitlin-cooling, line 20: And here too?

test-resources/headless-core-stub/lambdas/start/src/services/signing-service.ts
```diff
@@ -1,16 +1,65 @@
...
+    }
+
+    const jwks = await data.json();
+
+    if (!jwks.keys || !Array.isArray(jwks.keys) || jwks.keys.length === 0) {
+        throw new Error(`Invalid or empty JWKS response from ${criEncryptionJwksEndpoint}`);
+    }
+    logger.info({ message: "Successfully retrieved public encryption jwks endpoint", ...jwks });
+
+    cachedPublicKey = (await importJWK(jwks.keys[0], "RSA-OAEP-256")) as KeyLike;
```
Caitlin-cooling, line 59: Should we check that the type of key is encryption? Just looking at [Core's JWKS endpoint](https://api.identity.staging.account.gov.uk/.well-known/jwks.json) and there is different types of keys there

jkunle: I suppose, it doesn't hurt. I though sign keys are to be in DIDdocuments, but yeah 👍🏾

Caitlin-cooling (changes requested):
Changes look good, I know we have the fallback but I would prefer that we implement this behind a feature flag since we know for sure it's not going to work in kbv and check. 

If you want to test it all with address maybe you could just deploy the branch to that account? But would like to hold off merging this PR until we have a feature flag.

PR #480, opened 2025-05-21
OJ-3174: Move JWKS Endpoint from ENV variable to SSM Params

mikebeeby:
### What changed

Removed `PUBLIC_JWKS_ENDPOINT` from Session and Token env variables.
Added jwksEndpoint SSM params for multiple client ids.

### Why did it change

To support multiple client ids that may use different public JWKS endpoints, this moves the endpoint URL into SSM params per client.

### Issue tracking

- [OJ-3174](https://govukverify.atlassian.net/browse/OJ-3174)

[OJ-3174]: https://govukverify.atlassian.net/browse/OJ-3174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -213,49 +213,50 @@ Mappings:
       integration: false
       production: false
 
-  JwkEndpoint:
+  CoreJwksEndpoint:
+    Environment:
+      dev: https://api.identity.staging.account.gov.uk/.well-known/jwks.json
+      build: https://api.identity.staging.account.gov.uk/.well-known/jwks.json
+      staging: https://api.identity.staging.account.gov.uk/.well-known/jwks.json
+      integration: https://api.identity.integration.account.gov.uk/.well-known/jwks.json
+      production: https://api.identity.account.gov.uk/.well-known/jwks.json
+
+  StubJwksEndpoint:
     di-ipv-cri-address-api:
       dev: https://test-resources.review-a.dev.account.gov.uk/.well-known/jwks.json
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -1118,7 +1116,6 @@ Resources:
         Variables:
           POWERTOOLS_SERVICE_NAME: !Sub "${CriIdentifier}-access-token-2"
           ENV_VAR_FEATURE_CONSUME_PUBLIC_JWK: !FindInMap [ KeyRotationEnabled, !Ref CriIdentifier, !Ref Environment]
-          PUBLIC_JWKS_ENDPOINT: !FindInMap [ JwkEndpoint, !Ref CriIdentifier, !Ref Environment ]
```
Caitlin-cooling, line 1121: I wonder if we want to remove this in a separate PR once we've made the switch to using the SSM params?

mikebeeby: I can do, although I've changed the Mapping now and split it into Stub and Core. To get the same value for `PUBLIC_JWKS_ENDPOINT` I'd have to either put in the original mapping again or do some `if` logic to get it from Stub mapping if dev or build and Core mapping if staging?

Or we just leave it as is in the PR and accept it'll use the fallback like it mentions in the ticket?

Caitlin-cooling (approved):

PR #519, opened 2025-07-14
OJ-3183: Enable key rotation for CHECK HMRC API

jkunle:
## Proposed changes

As part of the Check HMRC  CRI encryption key exposure work, we need to ensure both versions of the Core Stub can fetch the public encryption key from the [/well-known/jwks.json ](https://api.review-hc.dev.account.gov.uk/.well-known/jwks.json) endpoint rather than using the hardcoded KMS logic. Currently, the Core Stubs configuration Key rotation for CHECK HMRC has been configured up to build along with Key rotation for headless core stub 

Simplified Integration Tests for Start Endpoint
- Stripped down start integration tests to focus only on HTTP response validation
- Removed JWT/JWE decryption and verification logic from integration tests
- Deleted helper files: jwe-decrypter.ts, jwt-verifier.ts, and kms.ts

### Why did it change

With key rotation flags enabled in the stub, JAR requests now use the JWKs endpoint from the CRI account rather than KMS core-infrastructure keys in the common account. This broke the original integration tests that relied on decrypting and verifying JWTs using keys that were previously available in the stub environment (Common account).

The decryption and verification aspects are will be move the Check HMRC CRI

This ongoing work that is Part of effectively managing keys i.e. to enable easier rotation of keys used by CRI clients such as core to make requests to CRI

### Issue tracking

- https://github.com/govuk-one-login/ipv-config/pull/1301
- [OJ-3183](https://govukverify.atlassian.net/browse/OJ-3183)

[OJ-3183]: https://govukverify.atlassian.net/browse/OJ-3183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


barnabycollins:
@jkunle - is this pipeline failure expected?

https://github.com/govuk-one-login/ipv-cri-common-lambdas/actions/runs/16290226717/job/45998530444?pr=519

PR #542, opened 2025-08-19
OJ-3254 - Enable key rotation feature flag in NINo check staging

barnabycollins:
### What changed

- Enabled key rotation feature flag in NINo check staging environment

### Why did it change

- We want to move towards having JWKS in use for encryption in all environments

### Issue tracking

OJ-3254

### Other considerations


Caitlin-cooling (changes requested):
Just checking you've deployed the latest version or and run the migration lambda so that the existing keys will be served on the jwks endpoint?

PR #698, opened 2026-05-11
OJ-3682: Implement headless CRI journey script

barnabycollins:
### What changed

- Configured npm to pull data-vocab from npm, not GitHub Packages
  - This involved a breaking version bump which included a new json-schema definition (2019-09) - we therefore need to [use a 2019-specific Ajv class](https://ajv.js.org/json-schema.html#draft-2019-09) to validate this in the headless core stub Start function
- Implemented headless CRI journey script

### Why did it change

- data-vocab is now published to npm; this makes it easier to consume as npm no longer needs to authenticate with GitHub in order to run an install
- The journey script should be useful for running simple tests against a deployed CRI (including localdev stacks, which currently often require manual pasting of values in the AWS dashboard)

### Issue tracking

- OJ-3682

### Other considerations


test-resources/headless-journey-script/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```sh
+# install correct node version (NB: requires node >=24 for running TS code natively)
+nvm install
+
+# install dependencies
+npm install
+
+# authenticate your AWS SSO profile
+aws sso login --profile your-aws-sso-profile
```
surajkumar: Is this really needed? A lot of people on the programme don't use AWS profiles and instead might use vault or authenticate their CLI with environment variables. I think you could probably remove all the credential logic like `credentials: fromSSO({ profile: input.awsProfile }),` and assume the CLI that the tool will be ran on is authenticated. 

The AWS libs should use automatically use environment vars so you could even do:

`AWS_PROFILE="<profile"> npm run journey`

surajkumar (dismissed):

test-resources/headless-journey-script/README.md
barnabycollins: Yeah, agree on this. Will adjust 👍

surajkumar (approved):

PR #700, opened 2026-05-12
OJ-3678: bump vulnerable ts libs

surajkumar:
### What changed
Ran `npm audit fix --force` to patch vulnerabilities.

Changes include bumping OTel deps as there was a vuln with `@opentelemetry/sdk-node` but in return bumped all the other OTel deps alongside for compatibility. 

### Issue tracking
- [OJ-3678](https://govukverify.atlassian.net/browse/OJ-3678)

[OJ-3678]: https://govukverify.atlassian.net/browse/OJ-3678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/package.json
```diff
@@ -37,7 +37,7 @@
         "@opentelemetry/instrumentation-net": "^0.43.0",
         "@opentelemetry/instrumentation-undici": "^0.10.0",
         "@opentelemetry/resource-detector-aws": "^1.11.0",
-        "@opentelemetry/sdk-node": "^0.57.1",
+        "@opentelemetry/sdk-node": "^0.217.0",
```
barnabycollins: Reckon it's worth bumping all of these to a recent version to ensure compatibility? Can't decide which is lower risk

surajkumar: Thanks - this ended up being needed.

lambdas/src/common/utils/otel-setup.ts
```diff
@@ -6,13 +6,8 @@ import { FsInstrumentation } from "@opentelemetry/instrumentation-fs";
...
 import { UndiciInstrumentation } from "@opentelemetry/instrumentation-undici";
-import { awsLambdaDetectorSync } from "@opentelemetry/resource-detector-aws";
-import {
-    envDetectorSync,
-    osDetectorSync,
-    processDetectorSync,
-    serviceInstanceIdDetectorSync,
-} from "@opentelemetry/resources";
+import { awsLambdaDetector } from "@opentelemetry/resource-detector-aws";
+import { envDetector, osDetector, processDetector, serviceInstanceIdDetector } from "@opentelemetry/resources";
```
barnabycollins, line 10: Sorry, do you mind explaining these changes? I can see they've been caused by the version bump but would be good to understand what led to this change - is it the same stuff but with a different name or is it a 'new way' of doing things?

surajkumar: These are from the `@opentelemetry/resources` and in v1.x the `Sync` variants existed to work around Lambda cold starts and other timing issues. Since we've now upgraded to the latest 2.7.1 - they were removed and consolidated to a single async API for all detectors. There's nothing different happening under the hood, these detectors are doing the same thing. 

Tldr: The sync variants were essentially a workaround that the maintainers decided was no longer needed once the SDK's startup sequencing was improved.

barnabycollins: Ahh, I hadn't seen the resource-detector-aws package had been bumped from v1 to v2. Makes sense, thanks :)

barnabycollins (approved):

PR #703, opened 2026-05-18
OJ-3693: Use latest ESLint and execute minor version bumps

barnabycollins:
## Proposed changes

NB: it may be easier to view these changes per-commit rather than as one big block.

These changes have been manually deployed to a localdev Check HMRC stack and I was able to complete a journey successfully.

### What changed

- Executed a version bump for all packages where a minor bump was available
- Bumped ESLint to latest version (10.4.0)
  - Replaced antiquated `@typescript-eslint/*` packages with newer recommended packages
  - Removed `eslint-plugin-prettier` which is really a non-recommended workaround for raising Prettier code style problems as linting problems
  - Updated and simplified ESLint config; migrated it to the modern flat config style
  - Hoisted shared dev dependencies to root package.json to avoid version conflicts and allow executing them in the root of the repo
  - Fixed issues raised by updated ESLint rules and config
- Stumbled across a number of repo config issues, many of which blocked this PR once the above was done:
  - `test-resources/package.json` wasn't actually defined as a workspace under the root `package.json`; only `package.json`s below the `test-resources` directory were. Therefore removed child `package.json`s in the test-resources stack and consolidated dependencies into `test-resources/package.json` (including unifying AWS SDK versions)
  - `mirrors-eslint` pre-commit hook doesn't seem to support new ESLint
  - Found that `@types/jest` was still installed and globals were enabled in vitest, helping to mask the presence of the Jest dependency. Disabled globals and removed `@types/jest`.
  - Removed a number of unused `tsconfig.eslint.json` and `tsconfig.json` files
  - Replaced `setEnvVars.js` files with `env: {...}` config in the relevant `vitest.config.ts` files
  - Removed `working-directory` defaults from GHA definitions and scoped them to only the necessary steps as these were causing `npm ci` to fail to install root-level dependencies in pre-merge checks

### Why did it change

- At least one of our packages was causing intermittent issues with lockfile corruption in Common Lambdas and OAuth Common, and we want to be on recent versions where possible anyway. A minor version bump resolved the issue.
- ESLint and related packages cause fairly regular vulnerabilities and are sometimes fiddly to fix; having them on the latest version should ensure fixes are faster and smoother
- Fixing these issues will result in an improvement to my mental health

### Issue tracking

- OJ-3693

### Other considerations


lambdas/src/handlers/create-auth-code-handler.ts
```diff
@@ -25,6 +25,12 @@ export class CreateAuthCodeLambda implements LambdaInterface {
     @logger.injectLambdaContext({ clearState: true })
     public async handler(
         event: APIGatewayProxyEvent,
+        /*
+            eslint-disable-next-line @typescript-eslint/no-unused-vars
+            --
+            The _context variable is unused but removing it changes the function's type to one that the
+            @metrics.logMetrics decorator doesn't include in its list of available overloads.
+        */
```
nlsteers: you could remove all these disable blocks by adding the following to the linter config

```
'@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
```

barnabycollins: Done, thanks :)

MarieseMikely (approved):
LGTM

PR #732, opened 2026-06-30
Remove unused publicSigningJwkBase64 param from headless-core-stub

JessWinterborne:
### What changed

Remove `publicSigningJwkBase64` SSM parameter fetch from headless-core-stub

### Why did it change

When using the headless core stub alongside oauth common, which doesn't create this parameter, the fetch fails.

### Other considerations

I had to make fixes to the package-test-resources workflow so that this change would get picked up:
- Update `source-dir` build cache key to include all source files
- Clear SAM build cache before building test-resources


barnabycollins (approved):

barnabycollins:
Tested in Check HMRC with no regression 👍

PR #735, opened 2026-07-08
OJ-1588: fix - add source account to common lambdas permissions

MarieseMikely:
### What changed

- Add `SourceAccount` to the common lambda permissions.

### Why did it change

Restricting access to lambdas to only those parts of the system that need access improves our security boundaries.

### Issue tracking

- [OJ-1588](https://govukverify.atlassian.net/browse/OJ-1588)

[OJ-1588]: https://govukverify.atlassian.net/browse/OJ-1588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -1583,6 +1583,7 @@ Resources:
       Action: lambda:InvokeFunction
       FunctionName: !GetAtt AccessTokenFunctionTS.Arn
       Principal: apigateway.amazonaws.com
+      SourceAccount: !Sub "${AWS::AccountId}"
 
   AccessTokenTSAliasPermission:
     Type: AWS::Lambda::Permission
@@ -1590,6 +1591,7 @@ Resources:
       Action: lambda:InvokeFunction
       FunctionName: !Ref AccessTokenFunctionTS.Alias
       Principal: apigateway.amazonaws.com
+      SourceAccount: !Sub "${AWS::AccountId}"
 
   AccessTokenFunctionLogGroup:
     Type: AWS::Logs::LogGroup
@@ -2245,6 +2247,7 @@ Resources:
       Action: lambda:InvokeFunction
       FunctionName: !GetAtt SessionFunction.Arn
       Principal: apigateway.amazonaws.com
+      SourceAccount: !Sub "${AWS::AccountId}"
 
   SessionFunctionAliasPermission:
     Type: AWS::Lambda::Permission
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -1579,17 +1579,23 @@ Resources:
 
   AccessTokenTSPermission:
     Type: AWS::Lambda::Permission
+    Condition: IsDevEnvironment
```
surajkumar: If we add this then the private API won't be able to execute the Lambda in non-dev?

infrastructure/lambda/template.yaml
```diff
@@ -1579,17 +1579,23 @@ Resources:
...
 
   AccessTokenTSAliasPermission:
     Type: AWS::Lambda::Permission
+    Condition: IsDevEnvironment
     Properties:
       Action: lambda:InvokeFunction
       FunctionName: !Ref AccessTokenFunctionTS.Alias
       Principal: apigateway.amazonaws.com
+      SourceAccount: !Sub "${AWS::AccountId}"
+      SourceArn: !Sub "arn:${AWS::Partition}:execute-api:${AWS::Region}:${AWS::AccountId}:${PreMergeDevOnlyApi}/*/*/*"
```
surajkumar: Same here, we don't include the private API?

surajkumar (changes requested):

infrastructure/lambda/template.yaml
```diff
@@ -1579,17 +1579,23 @@ Resources:
 
   AccessTokenTSPermission:
     Type: AWS::Lambda::Permission
+    Condition: IsDevEnvironment
     Properties:
       Action: lambda:InvokeFunction
       FunctionName: !GetAtt AccessTokenFunctionTS.Arn
       Principal: apigateway.amazonaws.com
+      SourceAccount: !Sub "${AWS::AccountId}"
+      SourceArn: !Sub "arn:${AWS::Partition}:execute-api:${AWS::Region}:${AWS::AccountId}:${PreMergeDevOnlyApi}/*/*/*"
```
barnabycollins: We also need to allow the CRI's private APIGW to invoke these Lambdas. I don't think the APIGW IDs are currently available in this template, and adding another mapping probably isn't desirable, so we might just have to stick to `SourceAccount` for common-lambdas which should mitigate most of the risk.

surajkumar (approved):

